PR #13, opened 2022-04-01
Workflows for basic pathogen repo CI

tsibley:
The goal is to establish a common setup for CI for all or at least most
of our pathogen build repos.

Includes a reusable workflow¹ which can be called/invoked by our
workflows in other repos and a "starter" workflow template² that
performs such a call, for copying into the pathogen repo (either
manually or via the web UI which supports these starter workflows).

These are both basic for now but appropriate for our "simple" pathogen
builds like zika and measles.  We can add workflow parameters and
conditionals as necessary in the future to support CI for more complex
builds, like ncov.  (It won't take much!)

Related to <https://github.com/nextstrain/zika/pull/16>.

¹ https://docs.github.com/en/actions/using-workflows/reusing-workflows
² https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -1,0 +1,38 @@
+# This workflow is intended to be called by workflows in our various pathogen
+# build repos.  See workflow-templates/pathogen-repo-ci.yaml (a "starter"
+# workflow) in this repo for an example of what the caller workflow looks like.
+name: CI
+
+on:
+  workflow_call:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v3
+        with:
+          # Consider parameterizing the Python version. -trs, 1 April 2022
+          python-version: "3.7"
+
+      - run: python3 -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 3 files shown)


tsibley:
See https://github.com/nextstrain/zika/pull/17 as an example of calling this reusable workflow. I don't know how to test the "starter" workflow via the GitHub UI until it's on master here, but I'm also not too worried about supporting the UI method of adding the workflow to a repo since it's easy to copy manually without any other setup.

j23414:
The [CI build output](https://github.com/nextstrain/zika/runs/5795260786?check_suite_focus=true) looks correct to me

victorlin (approved):
Exciting!

j23414 (approved):

tsibley:
Now that this is merged, here's what the starter workflow UI looks like:

image

Clicking the "Configure" button takes you to a [special version](https://github.com/nextstrain/measles/new/master?filename=.github%2Fworkflows%2Fpathogen-repo-ci.yml&workflow_template=pathogen-repo-ci) of the "commit a new file" form:

image

tsibley:
(If I could I would make the starter workflow end up as just `ci.yaml` in the destination repo, but I don't see how to make the button do that. Doesn't really matter, and I expect not to use the button most times anyway.)

PR #16, opened 2022-04-04
Upload auspice/, results/, and Snakemake logs from pathogen repo CI runs to artifact storage

tsibley:
Seems likely to be useful at times for debugging why a CI run failed.

### Testing
Based on #18 so I can see it run in this repo's CI before merge. [Update: …and glad I did, as it caught a typo!]

- [x] Successful artifact upload in CI ([example run](https://github.com/nextstrain/.github/actions/runs/2241384419), see "outputs" at bottom)

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -73,4 +73,6 @@ jobs:
           path: |
             auspice/
             results/
+            benchmarks/
+            logs/
             .snakemake/log/
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -66,3 +66,11 @@ jobs:
           fi
 
       - run: nextstrain build --docker . ${{ inputs.build-args }}
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: outputs
+          path: |
+            auspice/
+            results/
+            .snakemake/log/


.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -41,3 +67,11 @@ jobs:
...
 
       - run: nextstrain build --docker . ${{ inputs.build-args }}
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: outputs
+          path: |
+            auspice/
+            results/
+            .snakemake/log/
```
victorlin, line 77: I noticed for ncov workflow there are also `benchmarks/` and `logs/` directories generated, is that specific to ncov workflow or not useful here?

victorlin (approved):

.github/workflows/pathogen-repo-ci.yaml
tsibley: Their presence is workflow-dependent (more so than `auspice/` and `results/`) but the names are conventional when we use them. I think it's fine to include patterns here that don't always match; when I typo'd `.snakemake/logs/` (note the plural) nothing broke it just didn't upload the log files because nothing matched. That said, I'm not sure that `benchmarks/` is all that useful for CI runs, where I expect we're always using vastly-reduced example datasets with runtime characteristics that aren't representative. `logs/` is probably useful though, if it exists!

tsibley: Added both in 3cc214b.

tsibley:
Rebased onto master to resolve conflicts.

PR #17, opened 2022-04-27
Add a shared setup-nextstrain-cli action for our GitHub Actions workflows

tsibley:
Standardize and centralize how our workflows setup a working copy of
`nextstrain` when that's all they need.  This makes maintenance of the
setup much easier.

Investigating <https://github.com/nextstrain/cli/issues/171> made me
realize we could standardize this for basic workflows that require
`nextstrain`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,12 @@
+name: CI
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  test-actions/setup-nextstrain-cli:
+    name: test: actions/setup-nextstrain-cli
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./actions/setup-nextstrain-cli
--- a/actions/setup-nextstrain-cli/action.yaml
+++ b/actions/setup-nextstrain-cli/action.yaml
@@ -1,0 +1,53 @@
+name: Setup Nextstrain CLI
+description: >-
+  This GitHub Actions action is intended to be called by workflows in our other
+  repos when they need a working copy of the Nextstrain CLI (the `nextstrain`
+  program) installed and don't particularly care about the exact Python
+  environment in which its installed.
+  
+  For example, this is typically the case when the workflow's primary step
+  invokes `nextstrain build` to run a build in a container runtime (locally via
+  Docker or remotely via AWS Batch).  Workflows that are more complex (e.g.
+  have additional dependencies to install, care about Conda, etc.) should
+  probably continue to install the Nextstrain CLI themselves for now.
+
+inputs:
+  cli-version:
+    description: >-
+      Version of Nextstrain CLI to install, as a dependency version spec
+      understood by Pip (i.e. conforming to PEP-0508, e.g. >3.2 or ==3.2.4).
+      Defaults to none, which means the latest version available will be
+      installed.
+    type: string
+    default: ""
+    required: false
+
+  python-version:
+    description: >-
+      Version of Python to use for Nextstrain CLI, as a string understood by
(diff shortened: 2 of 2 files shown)


joverlee521 (approved):
We should update the [pathogen-repo-ci.yaml](https://github.com/nextstrain/.github/blob/master/.github/workflows/pathogen-repo-ci.yaml) to use this action as well.

tsibley:
@joverlee521 Oh, yes, I meant to but forgot to in all the running around between repos! Done.

tsibley:
Hmm, so using it for pathogen-repo-ci actually might be more complicated than that. Need to test to see. If testing is hard/takes time, I may cut it out of this PR, merge just the new action, and make a separate PR for the pathogen-repo-ci bits.

actions/setup-nextstrain-cli/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+  Docker or remotely via AWS Batch).  Workflows that are more complex (e.g.
+  have additional dependencies to install, care about Conda, etc.) should
+  probably continue to install the Nextstrain CLI themselves for now.
+
+inputs:
+  cli-version:
+    description: >-
+      Version of Nextstrain CLI to install, as a dependency version spec
+      understood by Pip (i.e. conforming to PEP-0508).  Defaults to none, which
+      means the latest version available will be installed.
```
victorlin, line 19: Not sure when this input would be used, but I could see myself naively specifying cli-version='3.2.4'` which would error.

```suggestion
  cli-version:
    description: >-
      Version of Nextstrain CLI to install, as a dependency version spec
      understood by Pip (i.e. conforming to PEP-0508) (e.g. >3.2 or ==3.2.4).  Defaults to none, which
      means the latest version available will be installed.
```

victorlin (changes requested):

actions/setup-nextstrain-cli/action.yaml
tsibley: Added and repushed.

victorlin (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  - push
+  - pull_request
+
+jobs:
+  test-actions/setup-nextstrain-cli:
+    name: test: actions/setup-nextstrain-cli
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./actions/setup-nextstrain-cli
```
victorlin, line 12: Throw a `run: nextstrain --version` step in here?

victorlin: Oh I guess it is already in the called workflow. But maybe good to check that at least `nextstrain` is available in this context.

tsibley: Ah, the action already does that.

tsibley:
FTR, I got the YAML syntax, workflow validity, and workflow steps messed up in `ci.yaml` (not the new action, the thing testing the new action) and fixed it with the following commits on master:

- acda08fea02854c01d5a3394ae307dda4783150e
- c6cf6bbc8065a1a3598d8359a10b51fc06d2421b
- 6da1f55

orz. May I never think, "oh it'll be fine, I don't need to test it" again.

.github/workflows/ci.yaml
tsibley: @victorlin Added `run: nextstrain version --verbose` after the action in https://github.com/nextstrain/.github/commit/226a9812f6944e1120b5d08149158cb0cca8252b. ([example output](https://github.com/nextstrain/.github/runs/6202728220?check_suite_focus=true)).

PR #22, opened 2022-05-23
Fix workflow-templates mixup between docs-ci and pathogen-repo-ci

tsibley:
The "with" parameter block was added to the wrong file in "Workflows for
Sphinx docs CI" (a65c1b2).

### Related issue(s)
Related to #20.


victorlin (approved):
Oops, thanks for fixing!

PR #23, opened 2022-05-23
pathogen-repo-ci: Allow caller to pass env vars

tsibley:
This can be useful to set variables respected by Augur, Nextstrain CLI,
etc.  Implemented ourselves as a workflow input since the usual job.env
block is not (yet?) supported by GitHub Actions for jobs which call a
reusable workflow.

The input value is a YAML-formatted string, which you can easily produce
by pretending you're writing normal nested YAML within a literal
multi-line block scalar (introduced by "|"):

```yaml
   with:
     env: |
       FOO: bar
       I_CANT_BELIEVE: "it's not YAML"
```

When (if?) the usual job.env block is supported for reusable workflows,
it will be a small transformation for each caller to use it.

### Testing
Tested in a separate GitHub Actions testbed, though I may have mucked up the syntax in this particular file. We'll see when I try to use it!

Update: Usage in https://github.com/nextstrain/monkeypox/pull/8 found a small place I mucked up, which is now fixed and that PR's CI ran successfully.

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -22,10 +22,42 @@ on:
         default: ${{ github.repository }}
         required: false
 
+      env:
+        description: >-
+          Additional environment variables to set before the build, as a string
+          containing YAML.  This is easily produced, for example, by pretending
+          you're writing normal nested YAML within a literal multi-line block
+          scalar (introduced by "|"):
+
+            with:
+              env: |
+                FOO: bar
+                I_CANT_BELIEVE: "it's not YAML"
+
+          Do not use for secrets!  Instead, pass them via GitHub Action's
+          dedicated secrets mechanism.
+        type: string
+        default: ""
+        required: false
+
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - if: inputs.env
+        name: Set environment variables
+        env:
+          env: ${{ inputs.env }}
+        run: >
+          echo "$env"
+          | yq --output-format json .
+          | jq --raw-output '
+              to_entries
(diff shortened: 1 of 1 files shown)


.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -22,10 +22,42 @@ on:
...
+      - if: inputs.env
+        name: Set environment variables
+        env:
+          env: ${{ inputs.env }}
+        run: >
+          echo "$env"
+          | yq --output-format json .
+          | jq --raw-output '
+              to_entries
+            | map("\(.key)<<__EOF__\n\(.value)\n__EOF__")
```
victorlin, line 56: _[non-blocking]_

Is this heredoc for variables? How'd you know it works for `$GITHUB_ENV`? I tried running it directly:

```sh
I_CANT_BELIEVE<<__EOF__
it's not YAML
__EOF__
# bash: I_CANT_BELIEVE: command not found
```

victorlin (approved):
Seems hacky but probably the best option for now!

.github/workflows/pathogen-repo-ci.yaml
tsibley: It's not standard Bash, but something that GitHub Actions implements, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings.

tsibley: Added a comment about this in [c90c1b1](https://github.com/nextstrain/.github/pull/23/commits/c90c1b18f47ac69c9a9ba115f4f1e5083beda810).

tsibley:
> Seems hacky but probably the best option for now!

Definitely hacky! but yeah…

PR #24, opened 2022-05-24
Update README

tsibley:
New CI test to keep us honest about keeping the README up to date, and an update to the README to make that new CI test pass.

### Testing
- [x] New CI test [fails before fix](https://github.com/nextstrain/.github/runs/6578842093?check_suite_focus=true); passes after fix


victorlin (approved):
CI on README 🤯

PR #28, opened 2022-11-08
docs-ci: Add pip install option

### LINKED ISSUE #21: docs-ci: Make the Conda setup optional
From https://github.com/nextstrain/.github/pull/20#pullrequestreview-960919426:

> Make the Conda setup optional. Not all our repos use Conda for the docs build, and it my experience it seems to result in slower docs builds than pip. (IIRC, this also means we'd want to make the `run.shell` bit conditional on `setup-miniconda`, as that's the thing that restores `-euo pipefail` to the shell invocation another way.)

victorlin:

Some docs projects use pip instead of conda to set up the environment.

1. Make environment-file input (Conda setup) optional.

2. Add a new optional input to install from pip.

3. Use one job per environment setup option.

Specifying none of the setup options will cause both jobs to be skipped, but I think that's appropriate user error.

### Related issue(s)

- Closes #21 

### Testing

- [x] Checks pass
- [x] Jobs run as expected ([evidence](https://github.com/nextstrain/.github/actions/runs/3624983828), though note that the warnings do not result in an error since https://github.com/nextstrain/augur/commit/b0b8abc48706a71cd8edf4d53fb6a3999e2d2ef3 has yet to be applied)

### DIFF
--- a/.github/workflows/docs-ci.yaml
+++ b/.github/workflows/docs-ci.yaml
@@ -23,9 +23,15 @@ on:
 
       environment-file:
         description: >-
-          Path to conda environment file (e.g. docs/conda.yml)
+          Path to conda environment file. (e.g. docs/conda.yml)
         type: string
-        required: true
+        required: false
+
+      pip-install-target:
+        description: >-
+          Pip install target. (e.g. local directory which contains setup.py, project URL)
+        type: string
+        required: false
 
       make-target:
         description: >-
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-ci.yaml
```diff
@@ -28,15 +34,24 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        shell: bash -l {0}
```
tsibley, line 33: Doesn't the actual `make …` step also need Conda activated (when Conda is in use)?

tsibley (changes requested):
Yay, glad to see this. My idea had been to make this workflow read the RTD config file and install what's configured there, which I still think would be nicer, but it's more complex.

.github/workflows/docs-ci.yaml
victorlin: Of course 🤦 I'll fix that. Also created #32 to catch this in CI.

victorlin: Fixed with [force-push](https://github.com/nextstrain/.github/compare/f1a8182ae488bc0aaf8e31e7f08c7b24e06498cd..79386fbfe9bf8702f54edb988e0f907fc6806697).

PR #30, opened 2022-11-10
Add our now-common workflow for syncing RTD redirects

tsibley:
First implemented as <https://github.com/nextstrain/docs.nextstrain.org/commit/35016b7e>.

### Testing

- [x] Manually test using the workflow in this branch from another repo
- [x] Checks pass

PR #31, opened 2022-11-10
sync-rtd-redirects: Perform dry runs when not on the default branch

tsibley:
This accomplishes two important things:

1. It protects against misconfigurations of the calling workflow's triggers, e.g. previously if the calling workflow included "pull_request" events then syncs would happen when they shouldn't have because triggers were trusted to be correct.

2. It makes this workflow useful for PRs by previewing the changes that will happen upon merge.

### Related issue(s)

#30 

### Testing

- [x] Checks pass
- [x] Manually test from calling workflow

### DIFF
--- a/.github/workflows/sync-rtd-redirects.yaml
+++ b/.github/workflows/sync-rtd-redirects.yaml
@@ -40,7 +40,15 @@ jobs:
       - name: Install readthedocs-cli
         run: python3 -m pip install readthedocs-cli
 
-      - name: Sync redirects
+      - if: github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+        id: wet-run
+        name: Sync redirects
         run: rtd projects "${{ inputs.project }}" redirects sync -f "${{ inputs.file }}" --wet-run
+        env:
+          RTD_TOKEN: ${{ secrets.RTD_TOKEN }}
+
+      - if: steps.wet-run.conclusion == 'skipped'
+        name: Sync redirects (DRY RUN)
+        run: rtd projects "${{ inputs.project }}" redirects sync -f "${{ inputs.file }}" --dry-run
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-rtd-redirects.yaml
```diff
@@ -40,7 +40,15 @@ jobs:
       - name: Install readthedocs-cli
         run: python3 -m pip install readthedocs-cli
 
-      - name: Sync redirects
+      - if: github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
```
victorlin, line 43:
```suggestion
      - if: github.ref_type == 'branch' && github.ref_name == github.event.repository.default_branch
```

This is what I used in [docker-base/ci.yml](https://github.com/nextstrain/docker-base/blob/c4df6aab877d3b213f3716069c76907b328ded0c/.github/workflows/ci.yml#L13).

According to the [`github` context doc](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), this should be equivalent and I think it expresses the "branch-only" condition more clearly.

victorlin (approved):

.github/workflows/sync-rtd-redirects.yaml
tsibley: Nod. I saw that when authoring this and looking at ways of specifying the condition, but I personally preferred the `github.ref == …` condition here as it felt clearer to me that "this runs on a single ref". That said, I might be in the minority in being familiar and comfortable with git's fully-qualified refs.

I (somewhat accidentally) merged this a few minutes ago before seeing the comment here, so I didn't make the change above, but if it seems worth it to you then please feel free to do so.

victorlin: I'm fine with keeping it as-is, this was non-blocking.

PR #33, opened 2022-11-30
sync-rtd-redirects: Skip when `RTD_TOKEN` is empty

victorlin:

RTD redirects can only be synced with a valid RTD_TOKEN. If it is an empty string, it must be invalid. Ideally, making the secret required would catch all invalid token errors, however an empty string still passes the requirement. This job-level condition is the next best thing.

When called with an empty token, the workflow will still be triggered as configured but the job will be skipped instead of failing due to an invalid RTD token.

### Testing

- [ ] Checks pass (but they don't test this reusable workflow)
- [ ] Skips job on a repo that doesn't have `RTD_TOKEN` available
- [ ] Runs job on a repo with `RTD_TOKEN` available

### DIFF
--- a/.github/workflows/sync-rtd-redirects.yaml
+++ b/.github/workflows/sync-rtd-redirects.yaml
@@ -29,6 +29,7 @@ on:
 
 jobs:
   sync:
+    if: secrets.RTD_TOKEN != ''
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


.github/workflows/sync-rtd-redirects.yaml
```diff
@@ -29,6 +35,7 @@ on:
 
 jobs:
   sync:
+    if: ${{ !(inputs.run-on-nextstrain-only) || github.repository_owner == 'nextstrain' }}
```
tsibley, line 38: Though I agree we only intend/expect this workflow to run on our repos, there's nothing specific to Nextstrain about it, and I wouldn't begrudge anyone from using it outside of Nextstrain (though I'd probably just make a copy myself). The `run-on-nextstrain-only` option as a bypass aligns with that too. So if the sole goal is to prevent errors on forks that aren't configured with a token, then what about skipping the job when `RTD_TOKEN` isn't defined/available instead of checking repo owner?

Separately, the reason `${{ }}` is "required" despite the docs saying it isn't is because the file still has to be valid YAML syntax. While strings in YAML can normally be unquoted, there are exceptions where quotes are required, such as when specific YAML metachars are present in the string. The leading `!... |` construct here in the condition is the culprit that requires quoting, as it's a type tag (`!...`) and block scalar start (`|`). So the `${{ }}` changes that by making the `!...` no longer a type tag, and thus the `|` no longer a block scalar start. The point is that using normal YAML quoted strings:

```yaml
if: "!(inputs.run-on-nextstrain-only) || github.repository_owner == 'nextstrain'"
```

works just as well as an unquoted `${{ ... }}`.

victorlin: > what about skipping the job when RTD_TOKEN isn't defined/available instead of checking repo owner?

This sounds better 👍 I'll rework this PR with that approach.

victorlin: I tried 56037a0a19a5f452a840bc93dea8af90e7d32f81 but it's not that simple... `secrets` can't be used in `if` 🫤 https://github.com/actions/runner/issues/520

So there's no way to condition on the job level, but there might be a way to conditionally skip the steps that use `RTD_TOKEN`. I'll try that.

tsibley: Ahh, hmm. The (un)availability of `secrets` in `job.if` is indeed [noted in the docs](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability).

tsibley:
Closing this as it stalled out and doesn't seem necessary. We can always re-open/re-visit it if we find ourselves hung up by missing/empty tokens.

PR #38, opened 2023-05-05
pathogen-repo-ci: Log in to docker.io and ghcr.io if possible

tsibley:
See commit messages.

Related-to: <https://github.com/nextstrain/docker-base/pull/148>

### Testing

- [x] Checks pass

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -44,6 +44,10 @@ on:
         default: ""
         required: false
 
+permissions:
+  contents: read
+  packages: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -60,6 +64,17 @@ jobs:
           registry: docker.io
           username: nextstrainbot
           password: ${{ secrets.DOCKER_TOKEN_PUBLIC_READ_ONLY }}
+        continue-on-error: true
+
+      # Log in, if possible, to ghcr.io which we use for staging images in
+      # nextstrain/docker-base.  The automatic GITHUB_TOKEN is restricted to
+      # read-only access by the "permissions:" block above.
+      - name: Log in to ghcr.io
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
         continue-on-error: true
 
       # Transforms the inputs.env *string* containing YAML like this:
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -48,6 +48,20 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      # Log in, if possible, to docker.io (Docker Hub), since authenticated
+      # requests get higher rate limits (e.g. for image pulls).  Our org-level
+      # secret DOCKER_TOKEN_PUBLIC_READ_ONLY is available to all our public
+      # repos on GitHub but only available here to this reusable workflow when
+      # called with "secrets: inherit".  On Docker Hub, the token is granted
+      # "public read-only" access.
+      - name: Log in to docker.io
+        uses: docker/login-action@v2
+        with:
+          registry: docker.io
+          username: nextstrainbot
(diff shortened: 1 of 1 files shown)


tsibley:
The annotations when one of the optional login steps fails are kinda annoying in the UI:

image

(screenshot from using this workflow in https://github.com/nextstrain/docker-base/pull/148)

joverlee521 (approved):
Too bad about the annoying error outputs...Would it be too hacky to check if `secrets.DOCKER_TOKEN_PUBLIC_READ_ONLY` exists in a separate step before "Log in to docker.io"?

tsibley:
Hmm, yeah, that would work.

We could also ditch `docker/login-action` and instead call `docker login` for both registries directly (and conditionally) from a single step (plus an `if: always()` cleanup step that did `docker logout` for each registry).

victorlin:
> We could also ditch `docker/login-action` and instead call `docker login`

Maybe this will help with https://github.com/nextstrain/docker-base/issues/131...

tsibley:
Pushed 906908a to conditionalize the docker.io login on the presence of the secret, to avoid the annoying warnings. \o/

It works [without the secret](https://github.com/nextstrain/.github/actions/runs/4918511402/jobs/8784942590#step:2:7):

Screenshot from 2023-05-08 11-51-29


and [with it](https://github.com/nextstrain/.github/actions/runs/4918520017/jobs/8784962132#step:2:7):

image

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -44,10 +44,47 @@ on:
...
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      # Log in, if possible, to docker.io (Docker Hub), since authenticated
+      # requests get higher rate limits (e.g. for image pulls).  Our org-level
+      # secret DOCKER_TOKEN_PUBLIC_READ_ONLY is available to all our public
+      # repos on GitHub but only available here to this reusable workflow when
+      # called with "secrets: inherit".  On Docker Hub, the token is granted
+      # "public read-only" access.
```
victorlin, line 60: Thanks for setting this up!

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -44,10 +44,47 @@ on:
...
+          tee -a "$GITHUB_OUTPUT" <<<available=${DOCKER_TOKEN_PUBLIC_READ_ONLY:+yes}
+
+      - if: steps.DOCKER_TOKEN_PUBLIC_READ_ONLY.outputs.available == 'yes'
+        name: Log in to docker.io
+        uses: docker/login-action@v2
+        with:
+          registry: docker.io
+          username: nextstrainbot
+          password: ${{ secrets.DOCKER_TOKEN_PUBLIC_READ_ONLY }}
+        continue-on-error: true
```
victorlin, line 75: Is this an alternative?

```suggestion
      - env:
          DOCKER_TOKEN_PUBLIC_READ_ONLY: ${{ secrets.DOCKER_TOKEN_PUBLIC_READ_ONLY }}
        if: env.DOCKER_TOKEN_PUBLIC_READ_ONLY != ''
        name: Log in to docker.io
        uses: docker/login-action@v2
        with:
          registry: docker.io
          username: nextstrainbot
          password: ${{ secrets.DOCKER_TOKEN_PUBLIC_READ_ONLY }}
        continue-on-error: true
```

tsibley: Hmm! Maybe? The `env` context is available in `jobs.<job-id>.steps.if`, but it's not clear to me if the step env is applied by that point or if it only includes the job and workflow env. We could test it!

tsibley: The docs don't seem to mention it, but the [suggestion](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) is to use job env not step env for `step.if`:

> Instead, consider setting secrets as job-level environment variables, then referencing the environment variables to conditionally run steps in the job.

tsibley: Heh, I have a test but can't test it until GitHub Actions stops having its outage. :upside_down_face:

tsibley: https://github.com/nextstrain/.github/pull/41

PR #39, opened 2023-05-05
Add actions/run-nextstrain-build-on-aws-batch

joverlee521:

Creates a custom action to run Nextstrain builds on AWS Batch so that we can centralize and standardize how our pathogen repos set up these builds. Motivated by the latest hiccup with the Snakemake upgrade that required fixes across multiple pathogen repos.¹

I decided to make it a composite action instead of a reusable workflow to allow for per-repo custom steps such as setting up trial configs for trial builds² and to be able to use the `GITHUB_ACTION_PATH` to access the shared write-envdir script.

¹ https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1681328249848399 
² https://github.com/nextstrain/ncov/blob/4c5fc8f484cb7310b1039f5f8e847dad34fe4774/.github/workflows/rebuild-gisaid.yml#L38-L48

### Testing

- [ ] Checks pass
- [ ] See [trial](https://github.com/nextstrain/seasonal-flu/actions/runs/4897723596) in seasonal flu

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Invoked by our GitHub Actions workflows, including the reusable workflows below.
   ([README](actions/setup-ssh/README.md))
 - [Setup debugger](actions/setup-debugger/action.yaml) for interactive debugging of workflow jobs
   ([README](actions/setup-debugger/README.md))
+- [Run Nextstrain build on AWS Batch](actions/run-nextstrain-build-on-aws-batch/action.yaml)
 
 See also GitHub's [documentation on creating custom actions](https://docs.github.com/en/actions/creating-actions/about-custom-actions).
 
@@ -81,6 +82,10 @@ Used to setup other repos.
   Only for use in private repositories!
   ([template](workflow-templates/debugging-runner.yaml),
   [properties](workflow-templates/debugging-runner.properties.json))
+
+- Run Nextstrain build on AWS Batch
+  ([template](workflow-templates/run-nextstrain-build-on-aws-batch.yaml),
+  [properties](workflow-templates/run-nextstrain-build-on-aws-batch.properties.json))
 
 See also GitHub's [documentation on starter workflows](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization).
 
--- a/actions/run-nextstrain-build-on-aws-batch/action.yaml
+++ b/actions/run-nextstrain-build-on-aws-batch/action.yaml
@@ -1,0 +1,68 @@
+name: Run Nextstrain build on AWS Batch
+description: >-
+  This GitHub Actions action is intended to be called by workflows in our other
+  repos when they need to run a nextstrain build on AWS Batch.
+
+inputs:
+  runtime-envvar-names:
+    description: >-
+      A list of environment variable names for environment variables that
+      should be passed to the build. The envvar names should be separated by
+      a space, e.g. "GITHUB_RUN_ID SLACK_TOKEN SLACK_CHANNELS"
+    type: string
+    required: false
+
+  cpus:
+    description: >-
+      Number of CPUs/cores/threads/jobs to utilize at once.
+      Will be passed to the `--cpus` option for `nextstrain build` and the
+      `--cores` option for Snakemake.
+    type: number
+    default: 4
+    required: false
+
+  memory:
+    description: >-
+      Amount of memory to make available to the build. Passed to the
+      `--memory` option for `nextstrain build`.
+    type: string
+    required: false
+
+  build-args:
+    description: >-
+      Additional command-line arguments to pass to `nextstrain build` after
+      the build directory (e.g. to Snakemake).
+    type: string
+    default: ""
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - uses: nextstrain/.github/actions/setup-nextstrain-cli@master
+
+    - name: Creating env.d
+      shell: bash
+      run: |
+        "$GITHUB_ACTION_PATH"/write-envdir env.d \
+          ${{ inputs.runtime-envvar-names }}
+
+    - name: Launch build on AWS Batch
+      shell: bash
+      run: |
+        nextstrain build \
+          --aws-batch \
+          --detach \
+          --no-download \
+          --cpus ${{ inputs.cpus }} \
+          --memory ${{ inputs.memory }} \
+          --exec env \
+          . \
+            envdir env.d snakemake \
+              --cores ${{ inputs.cpus }} \
+              ${{ inputs.build-args }} \
+        | tee >(tail -n4 >> "$GITHUB_STEP_SUMMARY")
+      env:
+        NEXTSTRAIN_DOCKER_IMAGE: ${{ env.NEXTSTRAIN_DOCKER_IMAGE }}
+        AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
--- a/actions/run-nextstrain-build-on-aws-batch/action.yaml
+++ b/actions/run-nextstrain-build-on-aws-batch/action.yaml
@@ -7,8 +7,15 @@ inputs:
   runtime-envvar-names:
     description: >-
       A list of environment variable names for environment variables that
-      should be passed to the build. The envvar names should be separated by
-      a space, e.g. "GITHUB_RUN_ID SLACK_TOKEN SLACK_CHANNELS"
+      should be passed to the build's runtime. The environment variables that
+      the Nextstrain CLI automatically forwards to the runtime do not need to be
+      listed here. See the list of environment variables that are auto-forwarded in
+      the Nextstrain CLI repo (https://github.com/nextstrain/cli/blob/master/nextstrain/cli/hostenv.py)
+
+      The environment variables can be any environment variable specified at
+      any level (workflow, job, or step) of the workflow using it.
+      The environment variable names should be separated by a space,
+      e.g. "GITHUB_RUN_ID SLACK_TOKEN SLACK_CHANNELS"
     type: string
     required: false
 
@@ -16,7 +23,8 @@ inputs:
     description: >-
       Number of CPUs/cores/threads/jobs to utilize at once.
       Will be passed to the `--cpus` option for `nextstrain build` and the
-      `--cores` option for Snakemake.
+      `--cores` option for Snakemake. See `nextstrain build` docs for more details:
+      https://docs.nextstrain.org/projects/cli/page/commands/build/
     type: number
     default: 4
     required: false
@@ -25,13 +33,15 @@ inputs:
     description: >-
       Amount of memory to make available to the build. Passed to the
       `--memory` option for `nextstrain build`.
+      See `nextstrain build` docs for more details:
+      https://docs.nextstrain.org/projects/cli/page/commands/build/
     type: string
     required: false
 
   build-args:
     description: >-
       Additional command-line arguments to pass to `nextstrain build` after
-      the build directory (e.g. to Snakemake).
+      the build directory (i.e. additional arguments for Snakemake).
     type: string
     default: ""
     required: false
--- a/workflow-templates/run-nextstrain-build-on-aws-batch.yaml
+++ b/workflow-templates/run-nextstrain-build-on-aws-batch.yaml
@@ -1,0 +1,28 @@
+name: Run Nextstrain build on AWS Batch
+
+on:
+  workflow_dispatch:
+    inputs:
+      image:
+        description: "Specific starter image to use for the build. Will override the default image for `nextstrain build`."
+        required: false
+        type: string
+
+jobs:
+  deploy-build-to-aws-batch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: nextstrain/.github/actions/run-nextstrain-build-on-aws-batch@master
+        with:
+          runtime-envvar-names: "GITHUB_RUN_ID PAT_GITHUB_DISPATCH SLACK_CHANNELS SLACK_TOKEN"
+        env:
+          NEXTSTRAIN_DOCKER_IMAGE: ${{ inputs.image }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          # Not auto-forwarded to builds, variable names must be passed via runtime-envvar-names
+          GITHUB_RUN_ID: ${{ github.run_id }}
+          PAT_GITHUB_DISPATCH: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
+          SLACK_CHANNELS: slack-channel-name
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+
--- a/actions/run-nextstrain-build-on-aws-batch/write-envdir
+++ b/actions/run-nextstrain-build-on-aws-batch/write-envdir
@@ -1,0 +1,22 @@
+#!/bin/bash
+# usage: write-envdir <envdir> [[var1 [var2 [var3 […]]]]
+#
+# Writes the current value for each environment variable name given into the
+# directory <envdir>, one file per variable.  Creates <envdir> if it doesn't
+# already exist.
+#
+# Originally written by @tsibley, copied unmodified from the ncov-ingest repo:
+# https://github.com/nextstrain/ncov-ingest/blob/2a8f5bb4419998d5e3e13c97ebd03a1780ea47c2/bin/write-envdir
+#
+set -eou pipefail
+
+dir="${1:?no envdir path}"
+shift
+
+mkdir -pv "$dir"
+cd "$dir"
+
+for name in "$@"; do
+    echo "${!name}" > "$name"
+    echo "Wrote $dir/$name"
+done
--- a/devel/check-readme
+++ b/devel/check-readme
@@ -48,7 +48,8 @@ files-to-ignore() {
         'images/*' \
         actions/setup-ssh/!(*.yaml|README.md) \
         actions/setup-debugger/!(*.yaml|README.md) \
-        actions/shellcheck/!(*.yaml|README.md)
+        actions/shellcheck/!(*.yaml|README.md) \
+        actions/run-nextstrain-build-on-aws-batch/!(*.yaml|README.md)
 }
 
 relative-links() {
--- a/workflow-templates/run-nextstrain-build-on-aws-batch.properties.json
+++ b/workflow-templates/run-nextstrain-build-on-aws-batch.properties.json
@@ -1,0 +1,4 @@
+{
+  "name": "Run Nextstrain build on AWS Batch",
+  "description": "Starter workflow for setting up Nextstrain builds that need to run on AWS Batch."
+}


actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+      Additional command-line arguments to pass to `nextstrain build` after
+      the build directory (e.g. to Snakemake).
+    type: string
+    default: ""
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - uses: nextstrain/.github/actions/setup-nextstrain-cli@master
```
corneliusroemer, line 42: Shall we explicitly set the Python version here? Or use default?

corneliusroemer:
This is amazing, thanks for tackling @joverlee521!

Could you show how to add custom steps? I've never used a custom action. How to do the trial run deploy? That would be nice to have somehow built in as well if it's possible.

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Not sure we need to explicitly set Python version since we are just using the Nextstrain CLI to launch the build on AWS. Maybe I'm being naive in thinking that the Python version wouldn't affect the `nextstrain build` command?

joverlee521:
> Could you show how to add custom steps? I've never used a custom action. 

I made a [wip commit](https://github.com/nextstrain/ncov/commit/349cf15180408d3f608bd47199ad6db045c1cb2d) of what this would look like for the ncov gisaid build workflow (untested). 

> How to do the trial run deploy? That would be nice to have somehow built in as well if it's possible.

I'm not sure how to build in a trial run as part of the action since the Snakemake configs are slightly different across the repos. Each repo is responsible for setting up its custom trial configs that can then be passed to this action via the `build-args` input.

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+      Number of CPUs/cores/threads/jobs to utilize at once.
+      Will be passed to the `--cpus` option for `nextstrain build` and the
+      `--cores` option for Snakemake.
+    type: number
+    default: 4
+    required: false
+
+  memory:
+    description: >-
+      Amount of memory to make available to the build. Passed to the
```
jameshadfield, line 26: We should specify the string format to use here, which presumably is "you must specify the units" as the help for `--memory` states "Units of b, kb, mb, gb, kib, mib, gib are supported."

jameshadfield: (`--aws-batch-memory` is different -- "Amount of memory in MiB to request for job")

jameshadfield: Alternatively, since we use c5 instances, should we be picking the maximum memory allocated for a machine with the specified CPUs? Or vice versa?

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
+name: Run Nextstrain build on AWS Batch
+description: >-
+  This GitHub Actions action is intended to be called by workflows in our other
+  repos when they need to run a nextstrain build on AWS Batch.
+
+inputs:
+  runtime-envvar-names:
+    description: >-
+      A list of environment variable names for environment variables that
```
jameshadfield, line 9: Where do these come from? Must they be a subset of the env variables set in the `deploy-build-to-aws-batch` below, or can you also supply your own when calling the rule?

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+      should be passed to the build. The envvar names should be separated by
+      a space, e.g. "GITHUB_RUN_ID SLACK_TOKEN SLACK_CHANNELS"
+    type: string
+    required: false
+
+  cpus:
+    description: >-
+      Number of CPUs/cores/threads/jobs to utilize at once.
```
jameshadfield, line 17: Similar to my comment about memory, is the recommendation that this be set to one of the [values of a c5 instance](https://aws.amazon.com/ec2/instance-types/) in order to not pay for CPUs which do nothing, or is that not really how instance provisioning works?

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+  memory:
+    description: >-
+      Amount of memory to make available to the build. Passed to the
+      `--memory` option for `nextstrain build`.
+    type: string
+    required: false
+
+  build-args:
+    description: >-
+      Additional command-line arguments to pass to `nextstrain build` after
```
jameshadfield, line 33: Is choosing these arguments identical to what you would provide to `nextstrain build` if it were calling snakemake for you (rather than us using `--exec` and then calling it ourselves? This came up recently so I think some docs here would be really helpful.

jameshadfield (commented):
Really good idea @joverlee521! All of my questions are just requests for more docs. I was tried to think about what questions I would have when I came to use this action some time in the future.

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: They can be _any_ environment variable you have set in the workflow when you are using this custom action. The `deploy-build-to-aws-batch` below is just an example template. In the [trial workflow](https://github.com/nextstrain/seasonal-flu/actions/runs/4897723596/workflow) I did for the seasonal-flu build, all three levels of env (workflow, job, step) were able to be passed to the build.

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Ah, I purposely left out any text here that would repeat the help output from the Nextstrain CLI to prevent them from getting out of sync. I could update the description with a link to the [nextstrain build docs](https://docs.nextstrain.org/projects/cli/en/stable/commands/build/).

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Ah, thanks for asking. I'll update the description to be clearer that these are additional arguments passed directly to Snakemake.

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: > Alternatively, since we use c5 instances, should we be picking the maximum memory allocated for a machine with the specified CPUs? Or vice versa?

Our compute environments set the [allocation strategy](https://docs.aws.amazon.com/batch/latest/userguide/allocation-strategies.html) to "BEST_FIT" so AWS will try to use an instance that fits the requested the resources at the lowest cost. I've tried to set a job to use 70GiB when the max memory is 72GiB, but AWS still used the next instance up (see [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1674254476788549)). 

So we just have to estimate the minimum resources needed for a job and AWS will figure out what instance to use.

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley: +1 for using the default. I think `python-version` will go away from `setup-nextstrain-cli` in a future update anyhow when it switches from installing with Pip to installing with our standalone installation.

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+          --exec env \
+          . \
+            envdir env.d snakemake \
+              --cores ${{ inputs.cpus }} \
+              ${{ inputs.build-args }} \
+        | tee >(tail -n4 >> "$GITHUB_STEP_SUMMARY")
+      env:
+        NEXTSTRAIN_DOCKER_IMAGE: ${{ env.NEXTSTRAIN_DOCKER_IMAGE }}
+        AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
```
tsibley, line 78: I don't think these are necessary? If the values are in the `env` context available to the job, they should be in the environment seen by the job's `run` block.

README.md
```diff
@@ -35,6 +35,7 @@ Invoked by our GitHub Actions workflows, including the reusable workflows below.
   ([README](actions/setup-ssh/README.md))
 - [Setup debugger](actions/setup-debugger/action.yaml) for interactive debugging of workflow jobs
   ([README](actions/setup-debugger/README.md))
+- [Run Nextstrain build on AWS Batch](actions/run-nextstrain-build-on-aws-batch/action.yaml)
```
tsibley, line 38: `nextstrain/.github/actions/run-nextstrain-build-on-aws-batch` seems a bit wordy and overly specific to a particular implementation. I think we ideally want an action that encapsulates a _concept_ ("do the builds"), not a specific implementation. We want to be able to evolve the internals of the action over time without always changing all its callsites.

What about `nextstrain/.github/actions/build` or `…/run-builds` or `…/pathogen-repo-builds` (akin to the existing `pathogen-repo-ci` shared workflow)?

(The name is an obvious place to attach my point here, but the point is broader and applies to the action at a whole and the interface it ideally presents.)

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley, line 52: Instead of `@master` I think we should use a fixed revision, for the same reasons as here:

https://github.com/nextstrain/.github/blob/fc841e889b68952c014e3d85bc022bfeec12d89a/.github/workflows/pathogen-repo-ci.yaml#L109-L130

workflow-templates/run-nextstrain-build-on-aws-batch.yaml
```diff
@@ -0,0 +1,28 @@
...
+        type: string
+
+jobs:
+  deploy-build-to-aws-batch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: nextstrain/.github/actions/run-nextstrain-build-on-aws-batch@master
+        with:
+          runtime-envvar-names: "GITHUB_RUN_ID PAT_GITHUB_DISPATCH SLACK_CHANNELS SLACK_TOKEN"
```
tsibley, line 18: _non-blocking_

One nicer way to format this in YAML would be:

```suggestion
          runtime-envvar-names: >-
            GITHUB_RUN_ID
            PAT_GITHUB_DISPATCH
            SLACK_CHANNELS
            SLACK_TOKEN
```

workflow-templates/run-nextstrain-build-on-aws-batch.yaml
```diff
@@ -0,0 +1,28 @@
...
+        description: "Specific starter image to use for the build. Will override the default image for `nextstrain build`."
+        required: false
+        type: string
+
+jobs:
+  deploy-build-to-aws-batch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: nextstrain/.github/actions/run-nextstrain-build-on-aws-batch@master
```
tsibley, line 16: _non-blocking_

Why `deploy-build-to-aws-batch` vs. `run-nextstrain-build-on-aws-batch`?

workflow-templates/run-nextstrain-build-on-aws-batch.yaml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          runtime-envvar-names: "GITHUB_RUN_ID PAT_GITHUB_DISPATCH SLACK_CHANNELS SLACK_TOKEN"
+        env:
+          NEXTSTRAIN_DOCKER_IMAGE: ${{ inputs.image }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          # Not auto-forwarded to builds, variable names must be passed via runtime-envvar-names
+          GITHUB_RUN_ID: ${{ github.run_id }}
```
tsibley, line 24: _non-blocking_

Why `GITHUB_RUN_ID`?

tsibley (changes requested):
Thanks for doing this work! I think this has the potential to become the foundation for broader, plug-and-play automation for our production pathogen builds.

I do kinda think that making this a composite action is more limiting than making it a reusable workflow, both in terms of longer-term implementation but also in how reusable workflows vs. actions manifest in the GitHub web UI. I realize that having `GITHUB_ACTION_PATH` eases access to `write-envdir`, but I do expect that to go poof at some point sooner than later with proper env var handling in Nextstrain CLI.

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+          --detach \
+          --no-download \
+          --cpus ${{ inputs.cpus }} \
+          --memory ${{ inputs.memory }} \
+          --exec env \
+          . \
+            envdir env.d snakemake \
+              --cores ${{ inputs.cpus }} \
+              ${{ inputs.build-args }} \
+        | tee >(tail -n4 >> "$GITHUB_STEP_SUMMARY")
```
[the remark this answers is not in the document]
joverlee521: Ah right, this is [GitHub Flavored Markdown](https://github.github.com/gfm/) so should be easy to create an indented code block. 

Hmm, I would also want to change the `/home/runner/work/seasonal-flu/seasonal-flu` directory to `.` to be able to directly copy/paste the command.

README.md
joverlee521: Hmm, I was wanting to be obvious that the build will be run on AWS Batch as opposed to within the GitHub Action itself. Now that I'm thinking about it, we can make the same workflow be able to run either way and just get toggled to AWS by a provided input.

actions/run-nextstrain-build-on-aws-batch/action.yaml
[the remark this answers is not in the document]
tsibley: You might consider only capturing the job id (e.g. as [ncov does](https://github.com/nextstrain/ncov/blob/4c5fc8f484cb7310b1039f5f8e847dad34fe4774/.github/workflows/rebuild-open.yml#L89)), and then interpolating that into your own Markdown template, a la what I did in conda-base ([interpolation](https://github.com/nextstrain/conda-base/blob/560e502589bf237eb66b94428f58e8b9528497a4/.github/workflows/ci.yaml#L108-L115), [template](https://github.com/nextstrain/conda-base/blob/560e502589bf237eb66b94428f58e8b9528497a4/.github/workflows/ci.release.md?plain=1)).

(Indentation is the OG code block, btw. Fences are the new kids on the (code) block.)

workflow-templates/run-nextstrain-build-on-aws-batch.yaml
joverlee521: Huh, I don't have a good answer here. Will consolidate to `run-build`

README.md
tsibley: Yeah, we should aim for the runtime to be ~transparent.

workflow-templates/run-nextstrain-build-on-aws-batch.yaml
joverlee521: In some of our existing builds, we have an `onstart` block that sends a Slack message that links to the GitHub Action run that started the build. These use the `GITHUB_RUN_ID` to construct the URL (e.g. [ncov-ingest/bin/notify-on-job-start)](https://github.com/nextstrain/ncov-ingest/blob/master/bin/notify-on-job-start)

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Huh, yeah that makes sense.

joverlee521:
> I do kinda think that making this a composite action is more limiting than making it a reusable workflow, both in terms of longer-term implementation but also in how reusable workflows vs. actions manifest in the GitHub web UI. 

I had the opposite thought that a composite action would be more flexible than a reusable workflow. However, I guess if we want this to be plug-and-play, it would make more sense to make this a standardized reusable workflow. 

> I realize that having `GITHUB_ACTION_PATH` eases access to `write-envdir`, but I do expect that to go poof at some point sooner than later with proper env var handling in Nextstrain CLI.

Hmm, maybe I can make `write-envdir` a composite action that gets called within the reusable workflow and we can eventually remove it. That will run into the same ref issue as `setup-nextstrain-cli`, but I think two hard-coded refs should be maintainable?

actions/run-nextstrain-build-on-aws-batch/action.yaml
corneliusroemer: What about versioning or tagging with dates to make it easier to remember?

actions/run-nextstrain-build-on-aws-batch/action.yaml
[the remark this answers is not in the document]
tsibley: Sure, but fencing would probably make ~zero difference here. ;-)

Fenced as `bash`:

```bash
nextstrain build --aws-batch --attach 2909dafd-02b3-476c-8d25-c0d95907ea57 .
```

Indented:

    nextstrain build --aws-batch --attach 2909dafd-02b3-476c-8d25-c0d95907ea57 .

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley: Hmm? There's no need to remember a specific revision's SHA. The revision used should be the latest at the time the calling workflow is introduced/needs updates. Am I misunderstanding your suggestion?

workflow-templates/run-nextstrain-build-on-aws-batch.yaml
tsibley: Ah. Hmm. That's a pretty distant coupling. I guess this is only a workflow template, but it seems likely to get copied/pasted as The New Incantation (similar to The Existing Incantation).

Ideally, in time, we have a better separation of concerns and move the notification stuff _out_ of the pathogen workflows themselves and into the automation (i.e. this action/workflow).

tsibley:
No need to wrap `write-envdir` in an action anyhow. You can checkout this repo (at whatever given revision) and run the program directly. An action would make that one step with a bunch of extra scaffolding for the action instead of two simpler ones using existing actions.

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+runs:
+  using: composite
+  steps:
+    - uses: nextstrain/.github/actions/setup-nextstrain-cli@master
+
+    - name: Creating env.d
+      shell: bash
+      run: |
+        "$GITHUB_ACTION_PATH"/write-envdir env.d \
+          ${{ inputs.runtime-envvar-names }}
```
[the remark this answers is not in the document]
joverlee521: > handling runtime-envvar-names fully as data and never as code

Could you clarify what you mean by this? I'm confused coming back to this now.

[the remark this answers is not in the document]
tsibley: Use quoting/escaping/etc. to make sure the value of `runtime-envvar-names` is always interpreted as a sequence of string literals instead of potentially as shell metachars, commands, etc.

This sort of quoting isn't enough, since it happens outside the shell:

```suggestion
          "${{ inputs.runtime-envvar-names }}"
```

Instead, passing via the environment is the first step:

```yaml
env:
  varnames: ${{ inputs.runtime-envvar-names }}
```

so that we can do quoting inside the shell:

```suggestion
          "$varnames"
```

which prevents the value being treated as code… but completely breaks the desired use of space-separated values.

To remedy that, we could use the same sort of [YAML-in-YAML workaround](https://github.com/nextstrain/.github/blob/cc6f4385a45bd6ed114ab4840416fd90cc46cd1b/.github/workflows/pathogen-repo-ci.yaml#L47-L64) I have in the pathogen-repo-ci workflow in this repo, e.g. so the input is specified as:

```yaml
- uses: nextstrain/.github/actions/run-nextstrain-build-on-aws-batch@master
  with:
    runtime-envvar-names: |
      - GITHUB_RUN_ID
      - SLACK_TOKEN
      - SLACK_CHANNELS
```

or instead use the simpler word splitting facilities of bash with `read`.

Putting that all together, e.g.:

```yaml
env:
  varnames: ${{ inputs.runtime-envvars-names }}
run: |
  read -a varnames -r -d $'\0' <<<"$varnames"

  "$GITHUB_ACTION_PATH"/write-envdir env.d "${varnames[@]}"
```

actions/run-nextstrain-build-on-aws-batch/action.yaml
jameshadfield: > Ah, I purposely left out any text here that would repeat the help output from the Nextstrain CLI to prevent them from getting out of sync. 

If I'm correctly imagining where this action will take us, it's plausible that the vast majority of AWS jobs we run will originate from this action. For instance, imagine we have _n_ caller workflows which each specify `memory: 32gib` (or similar), and then the nextstrain CLI changes its interface to wanting memory specified in kb as an INT; we'd now have to change _n_ caller workflows rather than just this one called workflow. Using major version tagging for this action, and coupling this action to a major version pin of nextstrain CLI, should avoid this fear of things getting out of sync. 

Put a different way, I think one should be able to go from running a snakemake workflow locally (without the nextstrain CLI) to using this action to automate the workflow without needing any specific knowledge of the nextstrain CLI itself. There's discussion above about "aim[ing] for the runtime to be ~transparent", and I think we should aim for the usage of nextstrain CLI itself to be ~transparent.

actions/run-nextstrain-build-on-aws-batch/action.yaml
jameshadfield: (Following up your comment from a different thread here, as that thread has moved to a slightly different focus)

> Our compute environments set the [allocation strategy](https://docs.aws.amazon.com/batch/latest/userguide/allocation-strategies.html) to "BEST_FIT" so AWS will try to use an instance that fits the requested the resources at the lowest cost. I've tried to set a job to use 70GiB when the max memory is 72GiB, but AWS still used the next instance up (see [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1674254476788549)).

> So we just have to estimate the minimum resources needed for a job and AWS will figure out what instance to use.

It would be nice to avoid the caller needing specific knowledge of AWS or our AWS config. I'm thinking here of providing a dropdown of preset memory/cpu combinations which maximise the efficiency of the machines the action uses (currently c5, and the presets would take into account the required headspace noted in that slack thread). If the caller really wants to use their own combination then they can specify them. 

Of course, this could also be done later on as an improvement to the action, it's not necessary from day one. But I would suggest pointing to the AWS c5 docs rather than the nextstrain CLI docs to start with. 

This somewhat couples the action to our AWS config + runner, but I'd think that any advancements there would be accompanied with a new major version release of this action. If the coupling doesn't happen here then it's implicitly happening in the caller actions which I think is worse for maintainability.

actions/run-nextstrain-build-on-aws-batch/action.yaml
jameshadfield: > I'll update the description to be clearer that these are additional arguments passed directly to Snakemake.

In a similar vein to my above comments, I would target this description to explaining how to go from running  `snakemake ...` to knowing what parts of the `...` I should include here. For instance, I should remove `--cores N`, and probably some other things right?

Perhaps it should also explain how to modify the arguments from `nextstrain build ...` to use here, as that's where uses may be coming from. In that case, however, isn't the runtime they are using important to know as it would affect the changes you need to make?

(I think no problem having verbose descriptions, in fact I prefer them!)

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,78 @@
...
+  build-args:
+    description: >-
+      Additional command-line arguments to pass to `nextstrain build` after
+      the build directory (i.e. additional arguments for Snakemake).
+    type: string
+    default: ""
+    required: false
+
+runs:
+  using: composite
```
jameshadfield, line 50: I was going to ask why you chose a composite action over a reusable workflow, but I see that's discussed in the main page of the PR!

I would also vote for this to be a reusable workflow. I can't think of a reason why you'd need this as a step inside a job, and I imagine many caller workflows will just call this single workflow (or action) and that's it!

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,78 @@
...
+description: >-
+  This GitHub Actions action is intended to be called by workflows in our other
+  repos when they need to run a nextstrain build on AWS Batch.
+
+inputs:
+  runtime-envvar-names:
+    description: >-
+      A list of environment variable names for environment variables that
+      should be passed to the build's runtime. The environment variables that
+      the Nextstrain CLI automatically forwards to the runtime do not need to be
```
jameshadfield, line 11: Assuming it's possible to list the environment variables passed to this called workflow via `${{ env }}` or `${{ inputs }}` or somesuch [1], a nicer interface would be for the caller to define the necessary env variables and the called workflow to forward everything to the runtime. The behavior of the nextstrain CLI would then become an implementation detail of this called workflow. For instance,

```yaml
- uses: nextstrain/.github/.../run-builds@v1
  with:
    - AWS_SECRET_ACCESS_KEY=...
    - AWS_ACCESS_KEY_ID=...
    - SOME_WORKFLOW_VAR=...
```

There's discussion above the runtime being ~transparent, and I think this interface makes it easy to imagine how the workflow could run the entire job within a GitHub action (i.e. not using AWS). 

[1] I'm somewhat writing this as if it were a reusable workflow not a composite action, but the idea should be applicable either way.

actions/run-nextstrain-build-on-aws-batch/action.yaml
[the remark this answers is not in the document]
jameshadfield: > Nice to use a step summary here

👍 Can we also provide direct URLs to the AWS console too, e.g. as [ncov does](https://github.com/nextstrain/ncov/blob/4c5fc8f484cb7310b1039f5f8e847dad34fe4774/.github/workflows/rebuild-open.yml#L91). Thanks!

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: (oh boy, I was confused by the term [transparent](https://en.wikipedia.org/wiki/Transparency_(human%E2%80%93computer_interaction)) for a bit there 😅 )

My original intention for this action/workflow is to reduce the work needed to _set up_ the GitHub Action workflow, but still be very clear that this is basically a wrapper around the Nextstrain CLI. I guess I'm not sure what is the benefit of adding another layer of abstraction on top of the Nextstrain CLI.

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: > Perhaps it should also explain how to modify the arguments from nextstrain build ... to use here, as that's where uses may be coming from.

Ah, this input is only for any additional Snakemake arguments, it does not pass arguments to Nextstrain CLI. I'll change the wording of the description to be less confusing.

jameshadfield: Yup - apologies - I could have been more precise. If I have a command running my workflow such as (any of):

```sh
snakemake --cores 1 -pf some/taget.x
nextstrain build someDirectory --resources A=B --profile C/D # what's my runtime? Does it matter?
```

I think the information required to map these to the `build-args` of this action should be provided by this action. However if the action is to be "basically a wrapper around the Nextstrain CLI" then "Additional command-line arguments to pass to `nextstrain build`" is probably enough, although I'd use the same language as `nexstrain` does: "Additional arguments to pass to the executed program".

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Ah, so pass any/all user defined environment variables to the runtime? 

In the reusable workflow, this would also apply to all `secrets` supplied by the caller. In #44, I'm not pre-defining any secrets in the reusable workflow and asking users to pass all secrets via `secrets: inherit`.

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: I was thinking of the [adhoc configs](https://github.com/nextstrain/ncov/blob/15817aa1376eded8d02f897988de55afdd109320/.github/workflows/rebuild-gisaid.yml#L38-L48) that we create in ncov workflows for trial runs, but you're right we don't need this to be a composite workflow. 

If anything, trial run configs can be set up in a separate job and the config can be passed via outputs.

actions/run-nextstrain-build-on-aws-batch/action.yaml
jameshadfield: [Notes after talking through this with Jover - but the notes are from my perspective!]

I think the caller should have to be explicit about what variables (inputs/secrets) are provided to the build runtime. While GitHub forces us to treat these differently, from the perspective of the build-runtime they're all just env variables. I'd suggest an interface where every build-runtime-needed variable must be defined explicitly by the caller workflow, and the called-workflow (this PR) just passes them through (or prunes out those which the CLI will automagically make available -- that's an implementation detail in my eyes). For instance:

```
- uses: nextstrain/.github/.../run-builds@v1
  with:
    - NCBI_EMAIL=...
  secrets:
    - SLACK_TOKEN=${{ secrets.SLACK_TOKEN }}
    - AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_ACCESS_KEY }}
    - AWS_ACCESS_KEY_ID=${{ secrets.AWS_ACCESS_KEY_ID }}
```

While this is more verbose than `secrets: inherit` I think it's much easier to understand given the number of moving parts here.

actions/run-nextstrain-build-on-aws-batch/action.yaml
jameshadfield: > this is basically a wrapper around the Nextstrain CLI

Fair enough - and this does have value - much better than having dozens of bespoke GitHub actions out there callling `--exec` and using envdir workarounds. I'd like to see this GitHub action become much more than just a wrapper, but this can happen in time. Assuming this PR remains a wrapper I'd still encourage the documentation ('description') for these inputs to be more self-explanatory rather than just pointing back to the CLI.

tsibley: @jameshadfield [wrote](https://github.com/nextstrain/.github/pull/39#discussion_r1206230420):

> Put a different way, I think one should be able to go from running a snakemake workflow locally (without the nextstrain CLI) to using this action to automate the workflow without needing any specific knowledge of the nextstrain CLI itself. There's discussion above about "aim[ing] for the runtime to be ~transparent", and I think we should aim for the usage of nextstrain CLI itself to be ~transparent.

Hmm. I don't understand (nor agree with, at least currently) this desire at all. What's your basis for it?

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley: @joverlee521 [wrote](https://github.com/nextstrain/.github/pull/39#discussion_r1210839259):

> Hmm, maybe we can add the dropdown as part of the workflow-template?

While that might be useful for initial creation of the workflow in a pathogen repo, note that the dropdown won't be available when it comes time to update the CPUs/memory needed. Those subsequent updates are likely to be necessary over time, as they've been several times for existing automated builds.

@jameshadfield [wrote](https://github.com/nextstrain/.github/pull/39#discussion_r1206239089):

> It would be nice to avoid the caller needing specific knowledge of AWS or our AWS config.
> […]
> This somewhat couples the action to our AWS config + runner, but I'd think that any advancements there would be accompanied with a new major version release of this action. If the coupling doesn't happen here then it's implicitly happening in the caller actions which I think is worse for maintainability.

I really don't see how presets vs. direct values reduces coupling for the calling workflow at all. It can't get away from talking about resources one way or another, and as long as those resources are on AWS, it's coupled to AWS.

The value of presets, as I see it, is not reducing coupling but making it easier to choose a resource combination that's most price-efficient for us.

If we wanted to actually reduce coupling, we'd have the calling workflow specify the arbitrary CPUs/memory it wants as a minimum and then have this action fit that into the most price-efficient-for-us CPUs/memory combination.

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley: @jameshadfield [wrote](https://github.com/nextstrain/.github/pull/39#discussion_r1210871142):

> […] then "Additional command-line arguments to pass to `nextstrain build`" is probably enough, although I'd use the same language as `nexstrain` does: "Additional arguments to pass to the executed program".

Hmm. I don't think using the exact same language makes sense out of context. "Additional arguments to pass to the executed program" makes sense in the context of the rest of the `nextstrain build --help[-all]` output¹, which also says things like "… will be run with snakemake" and the "Program to execute inside the runtime (default: snakemake)". I think different language is warranted in the different context here.

¹ All the context used to be part of just `--help`, but that changed when `--help-all` was added and the description of the additional args should probably change too.

actions/run-nextstrain-build-on-aws-batch/action.yaml
jameshadfield: This thread was motivated by me trying to push the PR towards my idea of an interface to running jobs on AWS across many repos. Jover wants this to be basically a wrapper around the Nextstrain CLI, which is fair enough. So these thoughts are no longer PR change requests, just my thoughts. With that out of the way:

> I really don't see how presets vs. direct values reduces coupling for the calling workflow at all. 

I was imagining ~3 parts coming together:
1. Guaranteeing that our current job queue uses c5 instances. If we change this, we'd create a new job queue (and deprecate the old one).
2. Explicitly using that job queue in this workflow, and if we introduce a new queue we'd release a new major version of this workflow (`@v2` etc)
3. This workflow would define resources via a choice of c5 instance, not memory/cpu combos. The job is going to be run on a c5 instance anyway, so why not centralise the mapping of cpu/memory allocations → c5 instance, including our best estimate of the memory overhead needed (as discussed in [this slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1674254476788549)). The alternative is that the caller workflows each make a guess of what memory goes with what CPU count (either a total guess, or an educated guess by reading the c5 docs).

> It can't get away from talking about resources one way or another, and as long as those resources are on AWS, it's coupled to AWS.

Yeah, the coupling will always be there, but I'm trying to avoid two situations: (1) where caller workflows make non-ideal choices of memory/cpus, with respect to the available instances and (2) if we change the instances our job queue uses then the caller workflows have no knowledge of that and will be using non-ideal resource combinations. 

> the calling workflow specify the arbitrary CPUs/memory it wants as a minimum and then have this action fit that into the most price-efficient-for-us CPUs/memory combination

That's another (good) option.

README.md
tsibley: In #44 the name is now `run-build`. I think this is an improvement! and don't want to belabor the point, but the more I think about it the more I think the name should parallel `pathogen-repo-ci` with a name like `pathogen-repo-build`.

I know I suggested `run-builds` (and even `build`), but they feel too generic esp. with what's already a very overloaded term.

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley: I see. To me, that's arguing for _tighter_ coupling! Doing that makes sense sometimes, and I see how it would if we're willing to make those guarantees about instance families, job queue versioning, action/workflow versioning, etc.

I'm wary of doing it that way though because those guarantees are extra overhead and though they mean we maintain price-efficiency even after an instance family/job queue switch, we still have to go update all the places (pathogen repos) using the previous versions. I don't think getting absolute price-efficiency is worth that time cost and operational overhead.

I'd rather do the alternative I proposed which _loosens_ coupling and still gets reasonable price-efficiency, because it doesn't require as much ongoing overhead and doesn't require updates to all the callers when we make changes.

> This thread was motivated by me trying to push the PR towards my idea of an interface to running jobs on AWS across many repos. Jover wants this to be basically a wrapper around the Nextstrain CLI, which is fair enough.

I think this will still be useful for running jobs on AWS across many repos…

I agree with Jover's goal here, and see this action/workflow as important "GitHub Actions glue". It reduces the barrier to go from running `nextstrain build` on your computer to running it under automation.

actions/run-nextstrain-build-on-aws-batch/action.yaml
tsibley: Agreed that `secrets: inherit` can be a bit opaque. This is not something the called workflow can enforce, though, so will be something we do (or don't do) by convention in the calling workflows.

tsibley: OTOH, `secrets: inherit` gives us more central control at the repo and org level without having to make modifications to all the callers.

actions/run-nextstrain-build-on-aws-batch/action.yaml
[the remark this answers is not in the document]
joverlee521: This has been addressed in #44

actions/run-nextstrain-build-on-aws-batch/action.yaml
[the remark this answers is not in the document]
joverlee521: addressed in #44

README.md
joverlee521: Addressed in #44

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Addressed with the `workflow-context` action in #44

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: Removed the `runtime-envvar-names` input and passing all explicit envvars and secrets to the build runtime in #44

actions/run-nextstrain-build-on-aws-batch/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+    - name: Launch build on AWS Batch
+      shell: bash
+      run: |
+        nextstrain build \
+          --aws-batch \
+          --detach \
+          --no-download \
+          --cpus ${{ inputs.cpus }} \
+          --memory ${{ inputs.memory }} \
```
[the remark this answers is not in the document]
joverlee521: Addressed in #44

actions/run-nextstrain-build-on-aws-batch/action.yaml
joverlee521: No longer relevant in #44 since reusable workflows do not get the `env` context from the calling workflow.

joverlee521:
Closing this in favor of #44.

PR #40, opened 2023-05-09
pathogen-repo-ci: Support a continue-on-error pass thru

tsibley:
Allows calling workflows to succeed even if the job in this called workflow fails.  This is a workaround for the calling working being unable to specify continue-on-error itself.¹

Related-to: <https://github.com/nextstrain/docker-base/pull/148>

¹ <https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow>

### Testing

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,6 +23,13 @@ jobs:
     with:
       repo: nextstrain/zika-tutorial
 
+  test-pathogen-repo-ci-failure:
+    uses: ./.github/workflows/pathogen-repo-ci.yaml
+    with:
+      repo: nextstrain/zika-tutorial
+      build-args: __BOGUS_BUILD_TARGET__
+      continue-on-error: true
+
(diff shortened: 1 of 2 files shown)


victorlin:
I wish the GitHub UI had some sort of indication on the commit level (2nd pic above, 1st arrow) that only continue-on-errors have failed. Mostly for this repo: since we expect `test-pathogen-repo-ci-failure` to always fail, the homepage will always show ❌ which needs an extra click to verify that just the one expected failure.

repo homepage

tsibley:
Yeah. We could remove the `test-pathogen-repo-ci-failure` job from this repo's CI if it gets in the way too much. It's really only testing that the GitHub Actions behaviour works as expected. That is, I don't think we need to continually test that our `continue-on-error: ${{ inputs.continue-on-error }}` line remains in working order.

PR #41, opened 2023-05-10
pathogen-repo-ci: Avoid the secret-check step before docker.io login

tsibley:
@victorlin pointed out¹ that we could launder access to the secrets context thru the step env context without the need for an extra step.

I avoid putting the secret itself into the environment (which only increases its potential visibility) and put only its presence/absence instead.

¹ <https://github.com/nextstrain/.github/pull/38#discussion_r1190281387>

### Testing

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,7 @@ jobs:
     uses: ./.github/workflows/pathogen-repo-ci.yaml
     with:
       repo: nextstrain/zika
+    secrets: inherit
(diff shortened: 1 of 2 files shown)


tsibley:
(Typo in branch name, but ah well, I'm not going to shuffle things around and create new PR noise just for that.)

victorlin (approved):
Nice, even better than my suggestion.

joverlee521 (approved):
✨

PR #47, opened 2023-08-30
Update PR template

victorlin:

A few small changes to improve the PR authoring and reviewing experience.

## Related issue(s)

N/A

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,8 @@
-### Description of proposed changes
+## Description of proposed changes
 
 <!-- What is the goal of this pull request? What does this pull request change? -->
 
-### Related issue(s)
+## Related issue(s)
 
 <!--
 Link any related issues here. Use GitHub's special keywords if appropriate¹.
@@ -11,7 +11,7 @@ Type `#` followed the name of an issue and GitHub will auto-suggest the issue nu
 ¹ https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests
 -->
 
-### Testing
+## Testing
 
 <!--
 Make sure checks are successful at the bottom of the PR.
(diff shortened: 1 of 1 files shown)


tsibley (commented):
I think we need to consider the question, "Who is the PR template for?"

If it's primarily for us, e.g. because we open the most PRs ourselves, then +1 for this change or similar changes as we should be streamlining the process of PR description usefulness for ourselves.

If it's primarily for external contributors, as I think it started out as during the pandemic for [ncov](https://github.com/nextstrain/ncov), then I'm not sure we want to ditch the hints and extra guidance.

victorlin:
@tsibley That's a good point. Since we welcome external contributions, I think the PR template should be for external contributors. It's easy for us to delete/ignore the hints that we don't need.

I've repurposed this PR to make a few smaller changes to the template.

.github/pull_request_template.md
victorlin, line 1: As an aside, I looked into whether we could use multiple PR templates. I was thinking there might be some sort of "template picker" such as the one for issues. While there is [support for multiple PR templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) (using the `.github/PULL_REQUEST_TEMPLATE` folder),  there's no "template picker" for PRs as of now. The only way to use templates in that folder is through [URL query parameters](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/using-query-parameters-to-create-a-pull-request) (e.g. `https://github.com/octo-org/octo-repo/compare/main...my-branch?quick_pull=1&template=issue_template.md`).

tsibley: Ah, yeah, I should have mentioned I also looked into that before commenting. Came to the same conclusion. Drats.

tsibley (approved):

PR #49, opened 2023-09-05
setup-nextstrain-cli: allow nextstrain update to continue-on-error

joverlee521:

I missed in https://github.com/nextstrain/.github/pull/48 that the command `nextstrain update` will exit with error for aws-batch and ambient runtimes. Instead of maintaining a list of runtime to run the update command, just allow it to continue-on-error.

I [found this error](https://github.com/nextstrain/forecasts-ncov/actions/runs/6089488339/job/16522370819#step:4:272) in separate work to update forecasts-ncov to use the pathogen-repo-build workflow with the aws-batch runtime.

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -221,7 +221,7 @@ jobs:
       # future changes to setup-nextstrain-cli.
       #
       #   -trs, 28 April 2022
-      - uses: nextstrain/.github/actions/setup-nextstrain-cli@e8e1f3213711d8c4b5022f7c29d3eb8ee3f46b4b
+      - uses: nextstrain/.github/actions/setup-nextstrain-cli@1fbb3f2e3bf29374d2706f720f65d4025853a89a
         with:
           runtime: ${{ matrix.runtime }}
           # Consider parameterizing the Python version. -trs, 1 April 2022
--- a/actions/setup-nextstrain-cli/action.yaml
+++ b/actions/setup-nextstrain-cli/action.yaml
@@ -67,6 +67,7 @@ runs:
 
     - run: nextstrain update
       shell: bash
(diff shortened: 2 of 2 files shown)


victorlin (approved):

joverlee521:
[Tested this in forecasts-ncov workflow](https://github.com/nextstrain/forecasts-ncov/actions/runs/6090170586)

Green checkmarks throughout the run even though the `nextstrain update` command failed. The Annotations section of the summary page does note the error, but the link to the log line is incorrect because of https://github.com/actions/runner/issues/1901

corneliusroemer:
I find that error in the annotations summary page confusing. Would it be possible to do something about that? Like add a note to the annotation: "this is an expected failure" or the like?
See: https://bedfordlab.slack.com/archives/C0K3GS3J8/p1695495002980659

victorlin:
I can't think of an easy way to hide or rewrite the error. I think the proper solution is to remove this `nextstrain update` step (it's only in place because the conda runtime does not use the latest version during `nextstrain setup`, as mentioned in #48).

corneliusroemer:
@victorlin wasn't the update step added to address the conda issue? We could fix the root cause there, or alternatively: make that setup step conditional on `conda` runtime being used?

@joverlee521 wrote:
>Instead of maintaining a list of runtime to run the update command, just allow it to continue-on-error.

In my view the confusing error message is worth maintaining a list of 1 that will need to almost never be updated 🙃

tsibley:
> wasn't the update step added to address the conda issue? We could fix the root cause there

The root cause was already fixed by https://github.com/nextstrain/cli/pull/312. I believe the `nextstrain update` is now extraneous.

victorlin:
@tsibley oh I missed that, thanks!

Issue #50, opened 2023-09-06
pathogen-repo-build: Display AWS Batch summary even when job was attached


joverlee521:
### Context
When running a build via the AWS Batch runtime as `--detached`, the nextstrain CLI outputs the command for reattaching to the build:

```
Detaching from job, as requested

Run the following command to re-attach to this job later to see output
and download results:

nextstrain build --aws-batch --attach dd3027c9-6094-4a60-ae57-8808ef6e7d34 /home/runner/work/seasonal-flu/seasonal-flu
```

The current step in the pathogen-repo-build workflow to check for the AWS Batch job id matches for this pattern since I was expecting AWS Batch jobs to always run detached. 

https://github.com/nextstrain/.github/blob/b59b79afffe3d15b8d49a106c6bc0de8ecd98d2c/.github/workflows/pathogen-repo-build.yaml#L215

However, we are moving towards [staying attached to AWS Batch builds](https://github.com/nextstrain/.github/pull/44#discussion_r1214646489). There's no similar message for when the job is run attached to AWS Batch, so we don't get the nice AWS Batch summary (e.g. [sequence counts workflow in forecasts-ncov](https://github.com/nextstrain/forecasts-ncov/actions/runs/6092447027)). If a user wants to attach to the same job locally, they would have to go digging in the GitHub Action logs again.

### Possible solution

Regardless of attachment status, the [nextstrain CLI always prints out the AWS Batch job id](https://github.com/nextstrain/cli/blob/3c93e87c2ac6bdd1b6a913070f709067b29a6cc1/nextstrain/cli/runner/aws_batch/__init__.py#L258) after the job is submitted.

joverlee521:
Could not figure out why `sed -nE 's/AWS Batch Job ID\: ([-a-f0-9]+)/\1/p'` wasn't working...then finally realized the text included [ANSI escape codes for formatting the text](https://github.com/nextstrain/cli/blob/3c93e87c2ac6bdd1b6a913070f709067b29a6cc1/nextstrain/cli/util.py#L45) 🤦‍♀️ 

`cat build.log | sed -nE 's/.+AWS Batch Job ID\:.+ ([-a-f0-9]+)/\1/p'` works as expected for me.

PR #51, opened 2023-09-06
pathogen-repo-build: update AWS Batch job id matching

joverlee521:

Previous pattern expected the AWS Batch job id to be in the example re-attaching command that is only output by the Nextstrain CLI when running the job with the `--detach` flag.

This checks for the AWS Batch job id in the stage output from the CLI immediately after the job was successfully submitted to AWS Batch.¹ This output exists regardless of the detachment status.

¹ https://github.com/nextstrain/cli/blob/3c93e87c2ac6bdd1b6a913070f709067b29a6cc1/nextstrain/cli/runner/aws_batch/__init__.py#L258

## Related issue(s)

Resolves https://github.com/nextstrain/.github/issues/50

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/pathogen-repo-build.yaml
+++ b/.github/workflows/pathogen-repo-build.yaml
@@ -212,7 +212,7 @@ jobs:
         name: Get AWS Batch job id
         id: aws-batch
         run: |
-          echo "AWS_BATCH_JOB_ID=$(tail -n1 build.log | sed -nE 's/.+attach ([-a-f0-9]+).+/\1/p')" >> "$GITHUB_ENV"
+          echo "AWS_BATCH_JOB_ID=$(sed -nE 's/.+AWS Batch Job ID\:.+ ([-a-f0-9]+)$/\1/p' < build.log)" >> "$GITHUB_ENV"
 
       - if: env.AWS_BATCH_JOB_ID
         name: Generate AWS Batch summary


.github/workflows/pathogen-repo-build.yaml
```diff
@@ -212,7 +212,7 @@ jobs:
         name: Get AWS Batch job id
         id: aws-batch
         run: |
-          echo "AWS_BATCH_JOB_ID=$(tail -n1 build.log | sed -nE 's/.+attach ([-a-f0-9]+).+/\1/p')" >> "$GITHUB_ENV"
+          echo "AWS_BATCH_JOB_ID=$(sed -nE 's/.+AWS Batch Job ID\:.+ ([-a-f0-9]+)$/\1/p' < build.log)" >> "$GITHUB_ENV"
```
tsibley, line 215: Does this pattern actually match the output? I think the ANSI escapes (e.g. for bold) would get in the way here with this pattern as-is.

tsibley (commented):
+1 for not relying on `--detach`. Adjusting the pattern makes sense for now.

Longer term, we could have Nextstrain CLI emit this data in a better way that's more suitable for programmatic usage (maybe piggybacking on https://github.com/nextstrain/cli/issues/113 or https://github.com/nextstrain/cli/issues/125). Would you be interested in doing that work?

.github/workflows/pathogen-repo-build.yaml
tsibley: Ahh, I see you [already ran into this](https://github.com/nextstrain/.github/issues/50#issuecomment-1709098826).

tsibley: I would think to future-proof this a bit more so it doesn't break when Nextstrain CLI eventually stops emitting ANSI control sequences to non-TTY stdouts.

tsibley (approved):

.github/workflows/pathogen-repo-build.yaml
joverlee521: Yup, leaving as-is as a work-around until we get a more suitable output from Nextstrain CLI. 

> Longer term, we could have Nextstrain CLI emit this data in a better way that's more suitable for programmatic usage (maybe piggybacking on https://github.com/nextstrain/cli/issues/113 or https://github.com/nextstrain/cli/issues/125). Would you be interested in doing that work?

Definitely interested in this work!

tsibley: Yay! Happy to chat more about it elsewhere, but some thoughts to get out of my head.

The data I'd think to emit would be the Batch job id, but also other info like the relevant S3 paths and maybe other bits and bots.

Some ways I've thought about emitting it in the past:

1. `nextstrain build` writes a JSON file when requested, e.g. by `--aws-batch-info job.json` (exact spelling TBD).
2. `nextstrain build` emits delimited YAML in its normal output, always, which can be easily extracted and parsed, e.g.

    Submitting job
    AWS Batch Job ID: 2ad078eb-630b-4ad2-80d0-bb898f5b6c48
    ---
    nextstrain-build-id: …
    aws-batch-job-id: …
    workdir-url: s3://…
    ...
    Watching job status

3. As part of https://github.com/nextstrain/cli/issues/113 and/or https://github.com/nextstrain/cli/issues/125, `nextstrain build` internally tracks job details and a new command provides subsequent access to them, e.g. `nextstrain build-info most-recent` or something.

(1) probably makes the most sense, especially as a starting point. (3) might be where we go eventually, but there's a lot more design considerations that we probably don't want/need to tackle now. (2) is kinda nice in some ways but annoying in others. And it's accomplishable with (1) anyway (well, assuming there's a YAML output option).

tsibley: This should probably become an issue over in the right repo…

tsibley: > I would think to future-proof this a bit more so it doesn't break when Nextstrain CLI eventually stops emitting ANSI control sequences to non-TTY stdouts. (But not required!)

https://github.com/nextstrain/.github/pull/126 orz

tsibley: > This should probably become an issue over in the right repo…

https://github.com/nextstrain/cli/issues/426

PR #53, opened 2023-09-14
pathogen-repo-build: Support assuming an AWS role via GitHub Actions' OIDC provider

tsibley:
The default role has a pretty limited set of permissions right now: basically only those required to submit and interact with AWS Batch jobs, but not, for example, permissions to upload to our production S3 buckets.  The permissions will probably need to be extended in the future as repos/builds opt into this workflow and this role assumption. The role's trust policy will also need to be extended to allow other repos to assume it.

While the role is currently manually managed, it should really be managed by a Terraform configuration (but one separate from the current nextstrain.org Terraform configurations).

Related-to: <https://github.com/nextstrain/private/issues/22>

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/pathogen-repo-build.yaml
+++ b/.github/workflows/pathogen-repo-build.yaml
@@ -29,13 +29,20 @@ on:
           Currently only supports docker, conda, and aws-batch.
           Defaults to "docker".
 
-          The aws-batch runtime requires the secrets:
+          The aws-batch runtime requires AWS credentials.  These may come
+          directly from secrets or indirectly from assuming a role via GitHub
+          Actions' OIDC provider.
+
+          The following secrets are used if present:
 
           - AWS_ACCESS_KEY_ID
           - AWS_SECRET_ACCESS_KEY
 
           They must be defined in the repo's Actions secrets and passed to this
           workflow with `secrets: inherit`.
+
+          If no secrets are present, the GitHubActionsRoleNextstrainBatchJobs
+          role is assumed (in both senses of the verb).
         type: string
         default: docker
         required: false
@@ -57,7 +64,13 @@ on:
 
           All environment variables provided via the env input and all secrets
           provided via `secrets: inherit` can be passed to the build runtime
-          via the `--env` option.
+          via the `--env` option.  If AWS credentials were acquired by the
+          GitHub Action job via role assumption, the following environment
+          variables are also available to be passed:
+
+            - AWS_ACCESS_KEY_ID
+            - AWS_SECRET_ACCESS_KEY
+            - AWS_SESSION_TOKEN
         type: string
         default: nextstrain build .
         required: false
@@ -136,11 +149,12 @@ on:
 env:
   NEXTSTRAIN_GITHUB_DIR: .git/nextstrain/.github
 
+permissions:
+  id-token: write
+
 jobs:
   workflow-context:
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
     steps:
       - id: workflow-context
         uses: nextstrain/.github/actions/workflow-context@master
@@ -167,16 +181,6 @@ jobs:
           ref: ${{ needs.workflow-context.outputs.sha }}
           path: ${{ env.NEXTSTRAIN_GITHUB_DIR }}
 
-      - name: Setup runtime ${{ inputs.runtime }}
-        uses: ./.git/nextstrain/.github/actions/setup-nextstrain-cli
-        with:
-          cli-version: ">=7.1.0"
-          runtime: ${{ inputs.runtime }}
-        # Required to set up aws-batch as default runtime
-        env:
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-
       - if: inputs.env
         name: Set environment variables
         env:
@@ -198,6 +202,21 @@ jobs:
           | jq 'del(.github_token)'
           | "$NEXTSTRAIN_GITHUB_DIR"/bin/json-to-envvars
           | tee -a "$GITHUB_ENV"
+
+      - if: inputs.runtime == 'aws-batch'
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-region: us-east-1
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-to-assume: ${{ secrets.AWS_ACCESS_KEY_ID == '' && 'arn:aws:iam::827581582529:role/GitHubActionsRoleNextstrainBatchJobs' || '' }}
+          credential-lifetime: 43200 # seconds, or 12 hours
+
+      - name: Setup runtime ${{ inputs.runtime }}
+        uses: ./.git/nextstrain/.github/actions/setup-nextstrain-cli
+        with:
+          cli-version: ">=7.1.0"
+          runtime: ${{ inputs.runtime }}
 
       - name: Run build via ${{ inputs.runtime }}
         env:


.github/workflows/pathogen-repo-build.yaml
```diff
@@ -199,6 +197,21 @@ jobs:
           | "$NEXTSTRAIN_GITHUB_DIR"/bin/json-to-envvars
           | tee -a "$GITHUB_ENV"
 
+      - if: inputs.runtime == 'aws-batch'
```
joverlee521, line 200: Would we want to support AWS roles for builds using docker/conda runtimes that upload/deploy to AWS S3? Or would those still be expected to supply their own credentials?

joverlee521: Wait, how would the temp credentials created even be passed to the build runtime?

joverlee521 (approved):
The changes here look reasonable to me! 

My main questions are about the permissions and management of the `GitHubActionsRoleNextstrainBatchJobs` role. 

> The default role has a pretty limited set of permissions right now: basically only those required to submit and interact with AWS Batch jobs, but not, for example, permissions to upload to our production S3 buckets.

Is the plan to manage upload permissions with another AWS Batch job role (as you [mentioned previously](https://github.com/nextstrain/private/issues/22#issuecomment-1593794969)). So the temp credentials here will allow the deployment and running of the Batch job, but permissions for AWS interactions within the build runtime will be managed by a separate role. 

> The role's trust policy will also need to be extended to allow other repos to assume it.

Is the plan to maintain a list of allowed repos in the role's trust policy and we would add repos to it as we add more pathogen workflows?

tsibley:
> Is the plan to manage upload permissions with another AWS Batch job role (as you [mentioned previously](https://github.com/nextstrain/private/issues/22#issuecomment-1593794969)). So the temp credentials here will allow the deployment and running of the Batch job, but permissions for AWS interactions within the build runtime will be managed by a separate role.

Unclear right now. A few options I see:

1. Extend the `GitHubActionsRoleNextstrainBatchJobs` role to allow uploading/downloading to S3 buckets used for pathogen builds (e.g. `nextstrain-data`, `nextstrain-data-private`, `nextstrain-ncov-private`, `nextstrain-staging`). Ensure the temporary credentials from role assumption are passed into the AWS Batch job via `nextstrain build`, as they are now.
2. Create a new role (or several) that allows the permissions above and have this workflow arrange to pass those temporary credentials into the Batch job (e.g. via `nextstrain build` + using `output-credentials: true` to `configure-aws-credentials`). This is a minor privilege separation and probably not worth it.
3. Update the default Batch job role (`NextstrainJobsRole`) to allow the permissions we need and stop passing credentials into the job via `nextstrain build`.
4. Create a new role (or several) that can be assumed by the Batch job itself and arrange via `nextstrain build` to have it assumed as part of the job submission.

> Is the plan to maintain a list of allowed repos in the role's trust policy and we would add repos to it as we add more pathogen workflows?

I'd probably allow all existing pathogen repos initially, even if they're not using role assumption right now, and then yeah, add more as we need. I'd think of this as "enabling" a repo to use our AWS Batch infra.

.github/workflows/pathogen-repo-build.yaml
tsibley: Yes, we would want to make it work for those runtimes on GitHub Actions too. It should be straightforward...

After `configure-aws-credentials`, the temp creds are in the environment as the standard `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` variables. To be available in the Nextstrain build runtime, they must then be either explicitly passed via `nextstrain build --env ...` or implicitly passed in the absence of any `--env` or `--envdir` option. So the caller of pathogen-repo-build does need to opt-into passing these if it's passing other env vars.

tsibley:
To choose amongst options, I'd think to consider:

1. Review permissions we currently grant via individual IAM users/policies for each pathogen that's automated
2. Decide if the union of those is palatable to grant widely to pathogen repos and any one with commit access to them
3. Decide if the union of those is palatable to grant to any one who can submit AWS Batch jobs in our group (review this) 
4. Determine if the maximum session lifetime of 12 hours for an assumed role via GitHub OIDC is sufficient for all our pathogen builds or not

Option 1 is tempting because it's ~transparent to pathogen builds and easy. But it maybe runs into session lifetime issues if we're passing those assumed temporary credentials into the Batch job.

Options 3 and 4 are tempting because they're also ~transparent and easy and won't run into session lifetime issues. However, it means being ok allowing everyone in the lab who can submit jobs to make all the changes our automated builds need to do.

.github/workflows/pathogen-repo-build.yaml
joverlee521: Ah okay, so this would require clear documentation for the user to know that they need to pass the `--env` options to interact with AWS in any way.

tsibley: Yep, the same as it is now.

joverlee521: Yeah, I guess I mean to document the exact environment variables that need to be passed. I think it's slightly less obvious here because the AWS environment variables are created within the GH Action workflow instead of being user provided secrets.

tsibley: Indeed, that's a good point.

.github/workflows/pathogen-repo-build.yaml
```diff
@@ -199,6 +197,21 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-region: us-east-1
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-to-assume: ${{ secrets.AWS_ACCESS_KEY_ID == '' && 'arn:aws:iam::827581582529:role/GitHubActionsRoleNextstrainBatchJobs' || '' }}
```
jameshadfield, line 206: Could you clarify what's going on here for me?

The `role-to-assume` line seems like it's assuming a role _if_ keys are not set as secrets. Seems sensible, as it allows users to provide their own IAM keys if the role isn't available / suitable to them.

But we also provide the keys here (`with.aws-access-key-id`, `with.aws-secret-access-key`), and from my reading of the [action docs](https://github.com/aws-actions/configure-aws-credentials/blob/main/action.yml) this means we  "want to assume a role using access keys rather than a web identity token". Perhaps if the keys are empty (which they are if we are going to assume a role) then it's as if we didn't provide them and so we'll assume a role via OIDC, but in this case why are they provided to this action at all?

tsibley: > Perhaps if the keys are empty (which they are if we are going to assume a role) then it's as if we didn't provide them and so we'll assume a role via OIDC…

Yes, this is what happens.

> …but in this case why are they provided to this action at all?

Because providing them to this action makes the secrets available in the environment for subsequent steps. This action is the single codepath to making AWS credentials available for the rest of the steps in the job, regardless of how those credentials were obtained.

jameshadfield: Gotcha. Thanks. [This table](https://github.com/aws-actions/configure-aws-credentials/tree/main#using-this-action) in the action's docs does imply that's what happens, but [the description of the inputs](https://github.com/aws-actions/configure-aws-credentials/blob/main/action.yml#L18-L20) doesn't say anything about that capability.

tsibley: Nod. I don't think GitHub Actions distinguishes between "empty" and "not set/not provided" inputs, so code consuming them doesn't either.

.github/workflows/pathogen-repo-build.yaml
tsibley: I'm going to address runtimes other than AWS Batch in another PR.

I did a quick look with GitHub search and all but two of our relevant workflows specify AWS Batch. The exceptions are [hepB](https://github.com/nextstrain/hepatitisB/blob/7bd1b05e55a7fe0179195d13d44abaf40755d1ef/.github/workflows/run-builds.yaml#L14-L17) and [seasonal-flu's fauna → S3 workflow](https://github.com/nextstrain/seasonal-flu/blob/873467c361738220041fcc0ff230def629c25150/.github/workflows/upload.yaml#L31-L44). They'd indeed both likely want conversion to role assumption in the future.

tsibley: [Added doc](https://github.com/nextstrain/.github/pull/53/files#diff-a708599ac70a96e634c67840937db538813ec036b702104dcf91f6e0e7f0849aR67-R73) on available AWS credentials.

tsibley: Made a checklist item on https://github.com/nextstrain/private/issues/98 for other runtimes.

tsibley:
This PR shouldn't have any functional change for our current usage, as I believe all callers provide explicit credentials themselves so the new role won't be assumed. As such, I'm inclined to merge it sooner than later to not leave it dangling. Further work to come will [extend the role's permissions](https://github.com/nextstrain/private/issues/95) (or otherwise provide for additional permissions) and ensure they're [passed in one way or another to the runtime](https://github.com/nextstrain/private/issues/96).

tsibley:
> While the role is currently manually managed, it should really be managed by a Terraform configuration (but one separate from the current nextstrain.org Terraform configurations).

Will be part of https://github.com/nextstrain/infra/compare/main...dev.

PR #54, opened 2023-09-14
pathogen-repo-build: Wait for AWS Batch jobs to finish

tsibley:
This avoids the disconnect between the success/failure status of a GitHub Actions workflow run and the actual AWS Batch job, which makes for easier reporting and debugging and generally less cognitive dissonance.

GitHub Actions workflow *runs* have a very high max timeout of 35 days, but each *job* in a workflow has a much lower max timeout of 6 hours. Many of our builds should be less than 6 hours¹, but here we support builds up to 24 hours by chaining together 4 GitHub Actions jobs. We can add more jobs to the chain if we need to, but I don't foresee that.

Nearly all of our builds are in public repos, which means they won't consume usage minutes from our GitHub Actions quota. However, they *will* consume concurrency limits from the quota. We were already frequently bumping into the default free-tier quota of 20 concurrent jobs, so adding more long-running jobs (to wait around for the AWS Batch jobs) was a nonstarter until we upgraded to a Team plan with its corresponding quota of 60 concurrent jobs.²

¹ As of 24 Jan, all Batch jobs except one in the prior week were sub 12 hours. The exception was the GISAID ncov-ingest job launched on 16 Jan. Next longest jobs were 10.5 hours, all GenBank ncov-ingest jobs. <https://github.com/nextstrain/private/issues/95#issuecomment-1909155013>

² <https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1705024616778389> <https://github.com/tsibley/blab-standup/blob/HEAD/2023-08-22.md>

---

See also the commit messages of supporting commits leading up to the final commit (above).

## Checklist

- [x] Checks pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+SHELL := /bin/bash -euo pipefail
+.SILENT:
+
+workflows := .github/workflows/*.yaml.in
+
+inputs  := $(sort $(wildcard $(workflows)) $(shell git ls-files --cached -- '$(workflows)'))
+outputs := $(inputs:.in=)
+
+## Regenerate all the things.
+all: $(outputs)
+
+## Regenerate all the things and error if anything changed.
+check: $(outputs)
+	git diff --exit-code --text HEAD -- $(outputs)
+
+## Regenerate an exploded workflow YAML.
+.github/workflows/%.yaml: .github/workflows/%.yaml.in PHONY
+	./devel/regenerate-workflow $<
+
+## Print this help message.
+help:
+	@perl -ne 'print if /^## / ... s/^(?<!##)(.+?):.*/make \1\n/ and not /^#( |$$)/' Makefile
+
+.PHONY: PHONY all check help
+PHONY:
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# This is a large generated file that, while text, it is not useful to
+# routinely show the diff of.  A diff can be forced as needed, e.g. with `git
+# diff --text`.
+/.github/workflows/pathogen-repo-build.yaml -diff
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,5 +74,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - run: make check
       - run: ./devel/check-readme
       - uses: ./actions/shellcheck
--- a/.github/workflows/pathogen-repo-build.yaml
+++ b/.github/workflows/pathogen-repo-build.yaml
@@ -1,8 +1,9 @@
+# DO NOT EDIT - GENERATED
+
 # This workflow is intended to be called by workflows in our various pathogen
 # build repos.  See workflow-templates/pathogen-repo-builds.yaml (a "starter"
 # workflow) in this repo for an example of what the caller workflow looks like.
 name: Pathogen repo build
-
 defaults:
   run:
     # This is the same as GitHub Action's `bash` keyword as of 20 June 2023:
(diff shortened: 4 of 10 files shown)


jameshadfield:
_I'll review this when I'm back for vacation, but no problem if it is merged before then._

Makefile
joverlee521, line 1: _Will continue reviewing the rest of PR tomorrow_

When I try to run `make` locally, I get an error:
```
generating .github/workflows/pathogen-repo-build.yaml
/bin/bash: -c: line 1: syntax error: unexpected end of file
make: *** [.github/workflows/pathogen-repo-build.yaml] Error 2
```

Probably due to older version of Make on macOS (3.81)?

tsibley: Yeah, seems like it probably doesn't support `.ONESHELL`. I'll verify and amend a fix.

tsibley: Yep, I didn't see when it was added, but I confirmed that `.ONESHELL` is not supported by Make 3.81 (and not mentioned in its info doc pages).

tsibley: Also doesn't support `.SHELLFLAGS`. Fixed up in [5a85020](https://github.com/nextstrain/.github/pull/54/commits/5a85020f245bac4e04d923c3fcbb4a4e5d98bb4f), but now in testing I noticed another issue with `git ls-files` version. Will fix that up too.

tsibley: Ok, I ultimately choose to move the multi-line Bash out of the Makefile and make it a separate program. There, it was easier to add improved error/edge case handling and not worry about old Make versions.

joverlee521: Thanks for updating! I can run `make` successfully now.

tsibley:
[Repushed](https://github.com/nextstrain/.github/compare/b8383d3da1a645c9d3a23429b8765b3399eb797e..57eb5e53e628782f9470db2d262e4d5ee1622eae) to fix a linting issue (but not bug per se) that shellcheck in CI found.

joverlee521 (approved):
Very cool use of YAML anchors and references, I will need to spend more time to become familiar with it. I was able to successfully run `make` and use the pre-commit hook. 

The actual changes to the pathogen-repo-build workflow looks good by inspection. I think reasonable to test it out with the ncov-ingest workflows next? 

In our discussion yesterday about standardizing GH Actions for pathogen repos, @jameshadfield raised a good question about whether this will work with your [proposed pathogen-repo-automation workflow](https://github.com/tsibley/blab-standup/blob/master/2024-02-07.md)...My gut feeling is it will _not_ work, but will have to try it out.

tsibley:
I agree testing with ncov-ingest or something like that seems like a good next step. Next week, I'll merge this and arrange for that testing.

tsibley:
Hmm, actually, ncov-ingest and ncov aren't already using this `pathogen-repo-build.yaml` workflow. I don't really want to adapt them to use it right now as part of testing this.

Merging, as I suggested quickly last Friday, will cause [all callers of this workflow](https://github.com/search?q=org%3Anextstrain+%2Fpathogen-repo-build%5B.%5Dyaml%2F+%28path%3A*.yml+OR+path%3A*.yaml%29&type=code) to immediately start using the new version in this PR.

So either we can test in production by merging, fixing anything that crops up… which is probably fine but maybe also breaks the world.

Or I can temporarily modify some subset of callers to point to the workflow on this branch instead and start with more limited testing of real work loads. I'll do that.

tsibley:
[Extracted some data](https://gist.github.com/tsibley/2402348bcef12edab76b94413c2a3908) from calling workflows.

tsibley:
I [modified mpox to use the workflow from this branch](https://github.com/nextstrain/mpox/pull/235). It works for CI.

Then I triggered a manual (trial run) test: https://github.com/nextstrain/mpox/actions/runs/7879309419/job/21499345399

We'll also see what happens tomorrow after daily ingest is run.

tsibley:
[Ingest](https://github.com/nextstrain/mpox/actions/runs/7889341284?check_suite_focus=true) worked this morning.

tsibley:
Merging this today so it's used for all workflows tomorrow morning. Things look good with mpox, so let's go.

tsibley:
Changed the mpox workflows back: https://github.com/nextstrain/mpox/commit/f608152a24f08354af3fcf9af28c8bcbbc034768

tsibley:
Everything overnight looks good:

image

PR #55, opened 2023-09-14
setup-nextstrain-cli: Use a standalone installation instead of Pip/PyPI

tsibley:
Simplifies and speeds up installation. Aligns with our recommended installation used by many users and many of ourselves. Drops the baggage of setting up a Python install as a side-effect. Discussed on Slack a while back.¹

In order to continue supporting PEP 440/508 (i.e. Pip-style) version range constraints in the cli-version input, I added support for those to the https://nextstrain.org/cli/download/... endpoint.²

I've audited all existing callsites I could find, internally and externally, for transitive reliance on the setup-python step. None seemed to rely upon it, so it's now gone. See also the added comment about removing the python-version input.

¹ <https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1694627907115309>
² <https://github.com/nextstrain/nextstrain.org/pull/797>

# Todo

- [x] Audit all existing callers for transitive reliance on `setup-python`
- [x] Remove `python-version` and `setup-python`
- [x] Test in CI on other platforms? currently CI here only calls on Linux.
- [x] Extend nextstrain.org endpoints to support version range specs, [c.f.](https://gist.github.com/tsibley/7386e83651eb8b0956fa426d117da9ed) (https://github.com/nextstrain/nextstrain.org/pull/797)

# Checklist

- [x] Checks pass

### DIFF
--- a/actions/setup-nextstrain-cli/action.yaml
+++ b/actions/setup-nextstrain-cli/action.yaml
@@ -29,9 +29,12 @@ inputs:
   runtime:
     description: >-
       Nextstrain runtime to configure and set as the default, as a string
-      understood by `nextstrain setup`.  Defaults to "docker".
+      understood by `nextstrain setup`.  Defaults to "docker" on Linux and
+      "conda" on macOS (where, on GitHub Actions, Docker isn't supported).  On
+      Windows, no runtime is set up by default: GitHub Actions runners can't run
+      Linux containers via WSL2 and our Conda runtime doesn't support Windows.
     type: string
-    default: docker
+    default: ${{ runner.os == 'Linux' && 'docker' || runner.os == 'macOS' && 'conda' || '' }}
     required: false
 
   python-version:
@@ -60,7 +63,8 @@ runs:
     - run: nextstrain version
       shell: bash
 
-    - run: nextstrain setup --set-default "$runtime"
+    - if: inputs.runtime
+      run: nextstrain setup --set-default "$runtime"
       shell: bash
       env:
         runtime: ${{ inputs.runtime }}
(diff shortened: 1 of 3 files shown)


tsibley:
Going to merge this now pre-review so I can see it in action (har har) today and address anything that arises. If it's a big flop, we'll just revert! As always, I'd welcome post-merge review and feedback.

actions/setup-nextstrain-cli/action.yaml
```diff
@@ -29,38 +29,63 @@ inputs:
...
+  # XXX TODO: Remove this in coordination with existing callers.  Callers must
+  # stop specifying it before we remove it here in order to not break.  I've
+  # audited all of our internal usages¹ of this action to check that they don't
+  # rely on the setup-python call and prepped those usages for the removal of
+  # this input.  There are also many external usages², mostly existing in forks
+  # of our repos.
+  #   -trs, 13 Feb 2024
+  #
+  # ¹ <https://github.com/search?q=org%3Anextstrain+%2Fsetup-nextstrain-cli%40master%2F+%28path%3A*.yaml+OR+path%3A*.yml%29&type=code&p=1>
+  # ² <https://github.com/search?q=%28NOT+org%3Anextstrain%29+%2Fsetup-nextstrain-cli%40master%2F+%28path%3A*.yaml+OR+path%3A*.yml%29&type=code&p=1>
```
victorlin, line 49: It looks like all external usages are either forks or exact copies of one of our repos. I did a quick spot check and didn't find any that are actively using the GitHub workflows.

I don't think coordination with the external usages is feasible. I would be ok with dropping it now that internal usages are up to date. In the slim chance that this breaks external usage, they should be able to contact us and we can help them fix it.

victorlin (commented):
👍

actions/setup-nextstrain-cli/action.yaml
tsibley: Nod. I'd be fine with that, but didn't want to start something that might cause broader fallout while I'm out. But by all means you can start it… :-)

PR #57, opened 2023-09-25
feat(pathogen-repo-ci): allow configuration of workflow root

corneliusroemer:
Current `pathogen-repo-ci` seems to assume that we don't need to customize workdir in `nextstrain build .`, i.e. we can't set `nextstrain build foobar`.

Our nextstrain/pathogen-repo-template, however, puts the main workflow in a folder. So we should add that feature.

See https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1695664810319929 for discussion

### Testing

It works: https://github.com/nextstrain/monkeypox/actions/runs/6303654582/job/17113366921

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -63,6 +63,14 @@ on:
           - conda
         required: false
 
+      workflow-root:
+        description: >-
+          Path to the root of the Snakemake workflow filder to run CI on.
+          Defaults to the root of the workflow repo.
+        type: string
+        default: ${{ github.workspace }}
+        required: false
+
       artifact-name:
         description: >-
           Name to use for build results artifact uploaded at the end of the
@@ -228,6 +236,7 @@ jobs:
           python-version: "3.7"
 
       - name: Copy example data
+        working-directory: ${{ inputs.workflow-root }}
         run: |
           if [[ -d example_data ]]; then
             mkdir -p data/
@@ -237,14 +246,17 @@ jobs:
           fi
 
       - run: nextstrain build . ${{ inputs.build-args }}
+        working-directory: ${{ inputs.workflow-root }}
 
       - if: always()
         uses: actions/upload-artifact@v3
+        env:
+          workdir: ${{ inputs.workflow-root }}
         with:
           name: ${{ inputs.artifact-name }}-${{ matrix.runtime }}
           path: |
-            auspice/
-            results/
-            benchmarks/
-            logs/
-            .snakemake/log/
+            $workdir/auspice/
+            $workdir/results/
+            $workdir/benchmarks/
+            $workdir/logs/
+            $workdir/.snakemake/log/
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -65,7 +65,7 @@ on:
 
       workflow-root:
         description: >-
-          Path to the root of the Snakemake workflow filder to run CI on.
+          Path to the root of the workflow folder (e.g. one that contains a Snakefile) to run CI on.
           Defaults to the root of the workflow repo.
         type: string
         default: ${{ github.workspace }}


.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -63,6 +63,14 @@ on:
           - conda
         required: false
 
+      workflow-root:
+        description: >-
+          Path to the root of the Snakemake workflow filder to run CI on.
```
victorlin, line 68:
```suggestion
          Path to the root of the workflow folder (e.g. one that contains a Snakefile) to run CI on.
```

1. Fix a typo
2. Use Snakemake as an example, not the assumed workflow engine

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -63,6 +63,14 @@ on:
           - conda
         required: false
 
+      workflow-root:
```
victorlin, line 66: "Workflow" is an overloaded term in this context (GitHub workflow vs. pathogen workflow). What about a simpler name like `path` or `directory`? That would look like:

```yaml
jobs:
  pathogen-ci:
    uses: nextstrain/.github/.github/workflows/pathogen-repo-ci.yaml@master
    with:
      path: phylogenetic/
```

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -237,14 +246,17 @@ jobs:
...
-            auspice/
-            results/
-            benchmarks/
-            logs/
-            .snakemake/log/
+            $workdir/auspice/
+            $workdir/results/
+            $workdir/benchmarks/
+            $workdir/logs/
+            $workdir/.snakemake/log/
```
victorlin, line 262: This currently only accepts paths without a trailing slash. I think it'd be good to accept both with and without, because even if we document support for only one in the input description, the error that happens here might not be very informative and only happens after everything else is done running.

I don't think there's an easy way to do something like Python's `os.path.join` inline here, so the solution might be to normalize as a first step before using the value.

victorlin (changes requested):

.github/workflows/pathogen-repo-ci.yaml
corneliusroemer: This sounds like a non-blocking feature request to me :)

.github/workflows/pathogen-repo-ci.yaml
corneliusroemer: Hmm, I'm not sure the simpler path is as clear. Yes there's ambiguity in `workflow-root` - but `path` is even more ambiguous.

.github/workflows/pathogen-repo-ci.yaml
victorlin: Yeah it works, so this can be non-blocking.

.github/workflows/pathogen-repo-ci.yaml
victorlin: Ok, also fine for this to be non-blocking. It works, I don't have any better suggestions, and this is something we can change easily if it's only used internally.

victorlin (approved):
Looks good given [this was tested successfully](https://github.com/nextstrain/monkeypox/actions/runs/6303654582/job/17113366921#step:1:39).

.github/workflows/pathogen-repo-ci.yaml
tsibley: `build-dir`, i.e. "the path for `nextstrain build`"

corneliusroemer: I'd be happy with `build-dir` or `build-root`

.github/workflows/pathogen-repo-ci.yaml
tsibley: > This sounds like a non-blocking feature request to me :)

This sounds like a flippant response to a reasonable request for robustness.

The trailing slash is not even a documented requirement in the input description. But even if it was, requiring a slash is an easily-avoided tripping hazard that we don't need to subject ourselves to.

tsibley, line 262: This doesn't work. The interpolation does not work like that. If you look at the "successful" test, [you'll see it didn't find any files](https://github.com/nextstrain/monkeypox/actions/runs/6303654582/job/17113366921#step:9:17) and [uploaded no artifacts](https://github.com/nextstrain/monkeypox/actions/runs/6303654582).

PR #62, opened 2023-09-25
Allow `pathogen-repo-ci` to run workflow in subdir via `build-dir` arg

corneliusroemer:

I'm trialling out putting our main auspice-producing workflow in a folder of a pathogen repo (this is what's currently suggested by our pathogen-repo-template).

This is currently not supported by `pathogen-repo-ci` as there's an implicit assumption that the root of the workflow is in the root of the repo.

In this PR I add support for specifying the workflow directory through an argument passed to pathogen-repo-ci.

Changes are backwards compatible as the default is set to what we previously used: `.`

Prior attempt: https://github.com/nextstrain/.github/pull/57#discussion_r1336389847 this had a bug and so I reverted it.

Now the bug has been addressed. See test run: https://github.com/nextstrain/monkeypox/actions/runs/6330482320/job/17192978189#step:9:17

With the way the argument is used, the `build-dir` argument _must not_ end in trailing slash. This is mentioned in the documentation of the argument.

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -63,6 +63,16 @@ on:
           - conda
         required: false
 
+      build-dir:
+        description: >-
+          The directory used for `nextstrain build`.
+          If provided, example data should be in an `example_data` directory
+          relative to this directory.
+          Defaults to the root of the repository.
+        type: string
+        default: "."
+        required: false
+
       artifact-name:
         description: >-
           Name to use for build results artifact uploaded at the end of the
@@ -94,12 +104,20 @@ jobs:
       - id: inputs
         env:
           runtimes: ${{ inputs.runtimes }}
+          build_dir: ${{ inputs.build-dir }}
         shell: bash
         run: |
           runtimes="$(yq --output-format=json --indent=0 . <<<"$runtimes")"
           echo runtimes="$runtimes" | tee -a "$GITHUB_OUTPUT"
+
+          # Remove potential trailing slash and set default to current directory
+          build_dir="${build_dir%/}"
+          build_dir="${build_dir:-'.'}"
+          echo build_dir="$build_dir" | tee -a "$GITHUB_OUTPUT"
+
     outputs:
       runtimes: ${{ steps.inputs.outputs.runtimes }}
+      build_dir: ${{ steps.inputs.outputs.build_dir }}
 
   build:
     needs: configuration
@@ -228,6 +246,7 @@ jobs:
           python-version: "3.7"
 
       - name: Copy example data
+        working-directory: ${{ needs.configuration.outputs.build_dir }}
         run: |
           if [[ -d example_data ]]; then
             mkdir -p data/
@@ -237,14 +256,15 @@ jobs:
           fi
 
       - run: nextstrain build . ${{ inputs.build-args }}
+        working-directory: ${{ needs.configuration.outputs.build_dir }}
 
       - if: always()
         uses: actions/upload-artifact@v3
         with:
           name: ${{ inputs.artifact-name }}-${{ matrix.runtime }}
           path: |
-            auspice/
-            results/
-            benchmarks/
-            logs/
-            .snakemake/log/
+            ${{ needs.configuration.outputs.build_dir }}/auspice/
+            ${{ needs.configuration.outputs.build_dir }}/results/
+            ${{ needs.configuration.outputs.build_dir }}/benchmarks/
+            ${{ needs.configuration.outputs.build_dir }}/logs/
+            ${{ needs.configuration.outputs.build_dir }}/.snakemake/log/
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,6 +26,13 @@ jobs:
     with:
       repo: nextstrain/zika-tutorial
 
+  test-pathogen-repo-ci-with-build-dir:
+    uses: ./.github/workflows/pathogen-repo-ci.yaml
+    with:
+      repo: nextstrain/monkeypox
+      build-dir: phylogenetic
+      artifact-name: mpox-phylogenetic
+
   test-pathogen-repo-ci-failure:
     uses: ./.github/workflows/pathogen-repo-ci.yaml
     with:


corneliusroemer:
Struggling to figure out how to deal with various `./` path issues so giving up for now and reverting #57

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -63,6 +63,16 @@ on:
           - conda
         required: false
 
+      build-dir:
+        description: >-
+          The `BUILD_DIR` for `nextstrain build $BUILD_DIR`.
+          Example data is expected to be in `$BUILD_DIR/example_data/`
+          Defaults to the root of the workflow repo.
+          Directory string must not end in a slash.
```
victorlin, line 71: _(non-blocking)_

Some minor wording changes. Notably, I'd avoid using an example variable name. If you want to template things, I'd suggest using the input name as-is (e.g. `nextstrain build <build-dir>`).

```suggestion
          The directory used for `nextstrain build `.
          If provided, example data should be in a `example_data` directory relative to this directory.
          Defaults to the root of the repository.
          Directory string must not end in a slash.
```

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -237,14 +248,15 @@ jobs:
...
-            auspice/
-            results/
-            benchmarks/
-            logs/
-            .snakemake/log/
+            ${{ inputs.build-dir }}/auspice/
+            ${{ inputs.build-dir }}/results/
+            ${{ inputs.build-dir }}/benchmarks/
+            ${{ inputs.build-dir }}/logs/
+            ${{ inputs.build-dir }}/.snakemake/log/
```
victorlin, line 262: There's back-and-forth here among the 3 commits - I think they can be squashed into a single commit.

victorlin (approved):

.github/workflows/pathogen-repo-ci.yaml
corneliusroemer: For sure

joverlee521:
I think the updates to the `pathogen-repo-ci` here is really grabbing for the flexibility that is allowed in `pathogen-repo-build`. 

The only crucial step that is missing from `pathogen-repo-build` is the copying of example data when available. 
We could add this step to individual pathogen repos by including Snakemake rules to copy over the example data as part of custom profiles. This is currently implemented in the [seasonal-flu repo's ci profile](https://github.com/nextstrain/seasonal-flu/blob/a6a7c26a8807f6ba274ab4f94b9ba76647ccc016/profiles/ci/prepare_data.smk) and gives the advantage of allowing users to [run the example build](https://github.com/nextstrain/seasonal-flu#example-build) without having to manually copy the data over. 

If we go with this pattern, we can slowly phase out the `pathogen-repo-ci` workflow and just maintain a single `pathogen-repo-build` workflow.

joverlee521:
I force-pushed updates to this PR and implemented the alternative approach of [using pathogen-repo-build in monkeypox](https://github.com/nextstrain/monkeypox/pull/214) so we can make direct comparisons of the two approaches. 

The `pathogen-repo-ci` takes care of more things behind the scenes so the workflow call is definitely much more succinct than the `pathogen-repo-build` workflow call. Is that enough to justify continuing to update/maintain both workflows?

tsibley:
> The `pathogen-repo-ci` takes care of more things behind the scenes so the workflow call is definitely much more succinct than the `pathogen-repo-build` workflow call. Is that enough to justify continuing to update/maintain both workflows?

One of the mismatches in the intents of `pathogen-repo-ci` and `pathogen-repo-build` is that the former expects to do more testing of runtimes while the latter does not. And that's what accounts for the verbose call to the latter when using it as a substitute for the former, e.g. needing a runtime matrix and threading that thru.

It's been nice in the past with `pathogen-repo-ci` to be able to update it to cover/test more bits (e.g. extending it to cover the Conda runtime too) without having to go update every call to it. Switching to `pathogen-repo-build` would lose that ability.

What if we keep `pathogen-repo-ci` but internally have it call `pathogen-repo-build`? Is that tractable?

joverlee521:
> What if we keep pathogen-repo-ci but internally have it call pathogen-repo-build? Is that tractable?

This seems reasonable. We would just have to thread the inputs through to the pathogen-repo-build. 
[Nesting reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#nesting-reusable-workflows) is possible, I just haven't tried it yet. Will look into it.

joverlee521:
> [Nesting reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#nesting-reusable-workflows) is possible, I just haven't tried it yet. Will look into it.

Tried this out in https://github.com/nextstrain/.github/pull/65

joverlee521:
> So maybe adopting your suggestion of this convention of a CI profile (at a typical path) + switching from pathogen-repo-ci → pathogen-repo-build is the way to go?

This made sense when I was thinking in the context of individual pathogen repo CIs. However, maybe not as nice for the CI jobs in docker-base/conda-base/(maybe augur?) if we have to add on to the matrix of inputs. 

We would still have to deal with the additional permissions for `pathogen-repo-build` if we ever tackle the [docker.io vs github.io registry issue](https://github.com/nextstrain/cli/issues/279) for docker-base. We would also have to deal with additional permissions for `pathogen-repo-ci` if we ever update it to use the `workflow-context` action. I also feel like things are up in the air because of the waffling on the standard file organization... 

Instead of making sweeping changes when we are still unsure about things, I'm fine with pausing on this PR. I'm going to just fix the [mpox repo CI with the pathogen-repo-build workflow ](https://github.com/nextstrain/mpox/pull/214) without affecting other repos.

joverlee521:
> This made sense when I was thinking in the context of individual pathogen repo CIs. However, maybe not as nice for the CI jobs in docker-base/conda-base/(maybe augur?) if we have to add on to the matrix of inputs.

Looping back to this after ~6 months: noodling on the idea to use `pathogen-repo-build` within pathogen CI workflows and add the `workflow_call` trigger to each CI workflow. Then the docker-base/conda-base CI workflows can just call the individual repo CI workflows. 

The one issue with this idea is that the `uses` call cannot be dynamic, so we wouldn't be able to directly use a matrix of repo names. There's a [workaround](https://github.com/orgs/community/discussions/45342#discussioncomment-6790679), but not sure if it's worth pursuing...

genehack:
Based on our lab meeting conversation today, I'm going to close this issue out in favor of a new one that I will be writing shortly.

PR #65, opened 2023-10-20
Nested pathogen-repo-ci

joverlee521:

Consolidates the pathogen-repo-ci and the pathogen-repo-build by having pathogen-repo-ci call on pathogen-repo-build. Suggested by @tsibley in https://github.com/nextstrain/.github/pull/62#issuecomment-1771362409. 

Technically the nested workflow is doable and it works (at least in [this repo's CI job](https://github.com/nextstrain/.github/actions/runs/6581545943)), but I'm not a big fan of the downsides of this approach. 

There are two major downsides:

1. We must hardcode the ref used for the pathogen-repo-build. We cannot
get around this with the workflow-context action because this is calling
a resuable workflow as a full job rather than an action step within a job.

2. We have to define additional permissions from the caller for both
the pathogen-repo-ci and the pathogen-repo-build with these updates
because if you specify the access for any of the scopes, all of those
that are not specified are set to none.

## Checklist

- [ ] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,20 +17,52 @@ jobs:
 
   test-pathogen-repo-ci:
     uses: ./.github/workflows/pathogen-repo-ci.yaml
+    permissions:
+      contents: read
+      packages: read
+      id-token: write
     with:
       repo: nextstrain/zika
     secrets: inherit
 
   test-pathogen-repo-ci-no-example-data:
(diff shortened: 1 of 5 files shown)


genehack:
Based on Jover's original description of the drawbacks with this approach and the lack of further movement in the ~6 months since, I think this is closable?

joverlee521:
Thanks for flagging @genehack! 

This is probably not the direction we'll be taking with pathogen CIs, so closing!

PR #75, opened 2024-02-16
setup-nextstrain-cli: Use Docker runtime on Windows with linux/amd64 containers

tsibley:
## Checklist

- [ ] Checks pass

### DIFF
--- a/actions/setup-nextstrain-cli/action.yaml
+++ b/actions/setup-nextstrain-cli/action.yaml
@@ -30,11 +30,10 @@ inputs:
     description: >-
       Nextstrain runtime to configure and set as the default, as a string
       understood by `nextstrain setup`.  Defaults to "docker" on Linux and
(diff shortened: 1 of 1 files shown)


tsibley:
Ah, yeah, [doesn't work](https://github.com/nextstrain/.github/actions/runs/7934187342/job/21664591321?pr=75#step:3:49):

```
image operating system "linux" cannot be used on this platform: operating system is not supported
```

I thought there was a chance, but I wasn't too hopeful because AIUI we need WSL2 support for Linux containers on Windows and GitHub Actions runners only enable WSL1.

victorlin:
I wondered if WSL2 support was on GitHub's roadmap at all - [apparently not](https://github.com/actions/runner-images/issues/5760#issuecomment-1158030908):

> [Jun 16, 2022] Per discussion with a team, unfortunately, there are no any plans to add WSLv2 support in the near feature.

Issue #58, opened 2023-09-25
CI shouldn't fail on expected failure


corneliusroemer:
### Current Behavior
We seem to fail CI despite expecting failure to be success in this run step:
https://github.com/nextstrain/.github/actions/runs/6163485323/job/16727132713

### Expected behavior
Overall CI result should be green checkmark if this fails. And be red if it succeeds actually

victorlin:
This was discussed when it was added in #40. I don't think there's a way around it, but also haven't looked very much.

genehack:
I'm looking at this again, and I'm probably lacking some sort of context, but what is this test even testing? A successful test happens when the `nextstrain build` command fails due to a bogus build argument, from what I can see — is there some common failure mode where a misconfigured `nextstrain` CLI returns success with a bogus build target? (I.e., what would this test _failing_ actually look like and would anybody notice?)

joverlee521:
> I'm looking at this again, and I'm probably lacking some sort of context, but what is this test even testing? 

This is testing the workaround for continue-on-error added in https://github.com/nextstrain/.github/pull/40 works as expected. Reading through that discussion again, @tsibley had [noted that we can eventually remove the test](https://github.com/nextstrain/.github/pull/40#issuecomment-1542675930). 

The workaround has been working for a ~year now, so I think it's safe to remove the test now!

genehack:
> The workaround has been working for a ~year now, so I think it's safe to remove the test now!

[Opened PR to remove](https://github.com/nextstrain/.github/pull/85)

PR #85, opened 2024-05-15
Remove intentional failure test #48 #58

genehack:
Per Jover,

> This is testing the workaround for continue-on-error added in #40
> works as expected. Reading through that discussion again, @tsibley
> had noted that we can eventually remove the test.

> The workaround has been working for a ~year now, so I think it's 
> safe to remove the test now!


joverlee521 (approved):

genehack:
:tada: Screenshot 2024-05-15 at 10 36 06

PR #87, opened 2024-05-22
write-envdir: Explicitly error with a better message when a requested var is unset

tsibley:
Eventually I'll dust off my WIP for `nextstrain write-envdir` which does this sort of nicety and more, and this Bash program will become obsolete.

## Checklist

- [x] Checks pass

### DIFF
--- a/bin/write-envdir
+++ b/bin/write-envdir
@@ -17,6 +17,21 @@ mkdir -pv "$dir"
 cd "$dir"
 
 for name in "$@"; do
+    # We could use [[ -v "$name" ]] if we didn't care about ancient Bash on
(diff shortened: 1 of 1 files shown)


tsibley:
[Testing this in actual context](https://github.com/nextstrain/zika/actions/runs/9198752365/job/25302147465) results in a better error message:

```
mkdir: created directory '.git/nextstrain/env.d'
Wrote .git/nextstrain/env.d/AWS_ACCESS_KEY_ID
Wrote .git/nextstrain/env.d/AWS_SECRET_ACCESS_KEY
error: AWS_SESSION_TOKEN is not set
```

victorlin (commented):
👍

PR #88, opened 2024-05-22
pathogen-repo-build: Only write AWS_SESSION_TOKEN into the envdir if it's not empty

tsibley:
No AWS_SESSION_TOKEN is present when callers provide static IAM user credentials via AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY. `write-envdir` requires all env vars given are defined.

## Checklist

- [x] Checks pass


tsibley:
[Tested ok](https://github.com/nextstrain/zika/actions/runs/9198774689/job/25302219824).

joverlee521:
Thanks for fixing 🙏

Issue #89, opened 2024-05-30
Update pathogen-repo-ci to "generic"/"smart" workflow


genehack:
### Context
Per [recent discussion](https://docs.google.com/document/d/1cdIwko_izJl5xuc1k54GDkP0Wt40eo6gFIkS5XQhCM4/edit), as well as various legacy conversations [here](https://github.com/nextstrain/pathogen-repo-guide/issues/25), [here](https://docs.google.com/document/d/1moFrcYo_gqQ8AT5tGHTpP0PwG-zMUaRIsERGcccRw-0/edit#heading=h.2o0ioyug4vc9), and [here](https://github.com/nextstrain/pathogen-repo-guide/issues/21), we want to update the `pathogen-repo-ci` GitHub Action to a "strict" version that can be used across all of our "modern" pathogen repos with zero configuration on the consuming repo side. 

### Description
Update the `pathogen-repo-ci` workflow so that:

1. _If_ the top-level `nextstrain-pathogen.yaml` file is present
2. For each build-name (i.e., each of `ingest`, `phylogenetic`, and `nextclade`):
    a. _If_ the repo contains a `<build-name>/Snakefile` file and a `<build-name>/build-configs/ci/config.yaml` file
    b. Run `nextstrain build <build-name>  --configfile build-configs/ci/config.yaml`

The consuming repo will be responsible for providing an appropriate config file that either sets up appropriate example data or results in a subset of a full workflow being run, so as to validate that the repo workflows have not been broken by a change. 

### Examples

This workflow, plus the appropriate build-config files, should be sufficient to enable CI for a pathogen repo:

```
name: CI
on:
  - push
  - workflow_dispatch
jobs:
  ci:
    name: CI
    uses: nextstrain/.github/workflows/pathogen-repo-ci.yaml
```

joverlee521:
100% support this direction. 

One additional check I would add to the `pathogen-repo-ci` is the presence of the top level `nextstrain-pathogen.yaml` file within the pathogen repo. This ensures `nextstrain build` is able to access of files between subdirectories, e.g. both `phylogenetic` might need to access files from the top level `shared` directory. ([nextstrain-cli v8.2.0 release notes](https://github.com/nextstrain/cli/releases/tag/8.2.0) has more details.)

tsibley:
+1 from me.

For the calling workflow in each repo, I'd extend the default set of triggers, e.g. something like

```yaml
on:
  push:
    branches:
      - main # or "master", as required
[...]

Issue #63, opened 2023-09-25
ENH: Allow pathogen-repo-ci to invoke workflow in subfolder


corneliusroemer:
### Context
Current pathogen-repo-ci seems to assume that we don't need to customize workdir in nextstrain build ., i.e. we can't set nextstrain build foobar.

Our nextstrain/pathogen-repo-template, however, puts the main workflow in a folder. So we should add that feature.

See https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1695664810319929 for discussion
### Description
A clear and concise description of what you want to happen

### Possible solution
Tried in #57 but introduce a bug in upload-artifact.
Tried to address in #62 but couldn't get it to work.

### Checklist

- [ ] Reenable monkeypox again for augur CI once this is resolved (disabled in https://github.com/nextstrain/augur/pull/1324)

victorlin:
[zika](https://github.com/nextstrain/zika) no longer has a top-level workflow, increasing the need for this. I'll make PRs to disable it from all CI for now.

joverlee521:
This has been superceded by https://github.com/nextstrain/.github/pull/90

PR #90, opened 2024-05-31
Update pathogen-repo-ci

genehack:

Adds "smarts" to `pathogen-repo-ci`. 

Also switches it over to using the "modern" style for building Nextstrain runtimes, using a sidecar repo to ensure everything runs from the same commit in the `nextstrain/.github` repo. 

## Related issue(s)

#89 #63 

## Checklist

- [x] Checks pass


actions/run-nextstrain-build-step/action.yaml
victorlin, line 1: (naming nitpicks)

Suggestion:  rename the input to `directory` to reflect the [name](https://docs.nextstrain.org/projects/cli/en/stable/commands/build/#cmdoption-nextstrain-build-arg-directory) in Nextstrain CLI. The word `step` in the action name is also redundant and can be removed.

```yaml
name: run-nextstrain-build
description: >-
  Runs a single `nextstrain build` command in a given sub-directory of
  a pathogen repo. Must be provided with the name of the
  sub-directory.

inputs:
  directory:
    description: The name of the sub-directory to run the build from
```

caller:

```yaml
- name: Run ingest
  uses: ./.git/nextstrain/.github/actions/run-nextstrain-build
  with:
    directory: ingest
```

genehack: I went back and forth on both of these and couldn't come up with something that felt right, so in the end I just pushed it trusting somebody would nitpick it for me – so thank you :grin:

joverlee521 (changes requested):
I appreciate using a composite action to keep the workflow DRY. I think it just needs some warnings that it can't be used as an independent step and requires the calling GH Action workflow to set up the Nextstrain CLI and build runtime.

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -196,7 +196,8 @@ jobs:
           '
           | tee -a "$GITHUB_ENV"
 
-      - uses: actions/checkout@v4
+      - name: Checkout repo
```
tsibley, line 221: > Add names to all steps

`actions/checkout` is so common that we've typically not named it, unless there are multiple checkouts/clones to disambiguate. Since that's the case here (in a subsequent commit), consider something like?

```suggestion
      - name: Checkout pathogen repo
```

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -226,7 +227,8 @@ jobs:
       #   approach. -trs ]
       #
       #   -trs, 28 April 2022
-      - uses: nextstrain/.github/actions/setup-nextstrain-cli@c1191de9d5e1a30e91d70b0fd1041d97ed1b2496
+      - name: Setup Nextstrain
```
tsibley: _nit nit nit_

```suggestion
      - name: Set up Nextstrain
```

(Yes, I realize the action path is `setup-nextstrain-cli` not `set-up-nextstrain-cli` but what can I say but _mea culpa_…)

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -201,47 +223,33 @@ jobs:
...
-      #   write) we decided not to update this workflow (yet?) to use the same
-      #   approach. -trs ]
-      #
-      #   -trs, 28 April 2022
-      - name: Setup Nextstrain
-        uses: nextstrain/.github/actions/setup-nextstrain-cli@c1191de9d5e1a30e91d70b0fd1041d97ed1b2496
+      - # Need to run this after the build repo is cloned so that cloning the
+        # build repo does not overwrite the .git dir and remove the extra support files
+        # that we need from nextstrain/.github repo
+        name: Checkout ${{ needs.workflow-context.outputs.repository }} (sha ${{ needs.workflow-context.outputs.sha }})
```
tsibley, line 229: _nit nit nit, follow prevailing style of comments in YAML_

```suggestion
      # Need to run this after the build repo is cloned so that cloning the
      # build repo does not overwrite the .git dir and remove the extra support files
      # that we need from nextstrain/.github repo
      - name: Checkout ${{ needs.workflow-context.outputs.repository }} (sha ${{ needs.workflow-context.outputs.sha }})
```

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -201,47 +223,33 @@ jobs:
...
+        uses: actions/checkout@v4
         with:
+          repository: ${{ needs.workflow-context.outputs.repository }}
+          ref: ${{ needs.workflow-context.outputs.sha }}
+          path: ${{ env.NEXTSTRAIN_GITHUB_DIR }}
+
+      - name: Setup Nextstrain runtime ${{ matrix.runtime }}
```
tsibley, line 236: _nit nit nit, percolated_

```suggestion
      - name: Set up Nextstrain runtime ${{ matrix.runtime }}
```

actions/run-nextstrain-build-step/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+            ! test -f $i && echo missing $i
+          done
+        fi
+      shell: bash
+
+    - id: upload-artifact
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{inputs.step }}-outputs-${{ matrix.runtime }}
```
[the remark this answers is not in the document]
tsibley: It is _very_ weird to me to have `matrix.runtime` as an implicit global var here way outside the calling context of a matrix job.

If this action is going to be so coupled to `pathogen-repo-ci.yaml`, what about moving it out of `actions/run-nextstrain-build/` and instead putting it in `.github/workflows/pathogen-repo-ci/run-nextstrain-build/` or similar?

actions/run-nextstrain-build/action.yaml
```diff
@@ -15,7 +15,7 @@ runs:
   steps:
     - id: run-build
       env:
-        STEP: ${{ inputs.step }}
+        STEP: ${{ inputs.directory }}
```
tsibley, line 18: Let's also not use `STEP`?

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -201,47 +223,33 @@ jobs:
...
+        uses: ./.git/nextstrain/.github/actions/run-nextstrain-build-step
+        with:
+          step: phylogenetic
 
-      - if: always()
-        uses: actions/upload-artifact@v4
+      - name: Run nextclade
+        uses: ./.git/nextstrain/.github/actions/run-nextstrain-build-step
         with:
-          name: ${{ inputs.artifact-name }}-${{ matrix.runtime }}
```
tsibley, line 247: `inputs.artifact-name` isn't being respected any more; it needs to be removed from inputs or (ideally, I think) threaded thru with a slightly changed meaning/usage. See also my other note about CI for `pathogen-repo-ci.yaml`.

actions/run-nextstrain-build-step/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+        else
+          echo "Skipping $STEP build due to one or more missing files."
+          for i in nextstrain-pathogen.yaml $STEP/Snakefile $STEP/build-configs/ci/config.yaml; do
+            ! test -f $i && echo missing $i
+          done
+        fi
+      shell: bash
+
+    - id: upload-artifact
+      if: always()
```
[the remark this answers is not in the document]
tsibley: The warnings are a lot. Note [this repo's CI run for this PR](https://github.com/nextstrain/.github/actions/runs/9325958578?pr=90) which has a ton.

tsibley (commented):
This repo's CI (`.github/workflows/ci.yaml`) is failing to actually test `pathogen-repo-ci.yaml` because it points at nextstrain/ebola and nextstrain/zika-tutorial, both of which result in all builds skipped.

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -201,47 +223,33 @@ jobs:
...
-      # future changes to setup-nextstrain-cli.
-      #
-      # [ Update 16 Feb 2024: We solved this for pathogen-repo-build.yaml, but
-      #   because it required a new permission on the GitHub tokens (id-token:
-      #   write) we decided not to update this workflow (yet?) to use the same
-      #   approach. -trs ]
-      #
-      #   -trs, 28 April 2022
-      - name: Setup Nextstrain
-        uses: nextstrain/.github/actions/setup-nextstrain-cli@c1191de9d5e1a30e91d70b0fd1041d97ed1b2496
```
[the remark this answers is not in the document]
genehack: Will remove!

actions/run-nextstrain-build-step/action.yaml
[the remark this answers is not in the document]
genehack: There's already this precedent: 

```
      - name: Setup Nextstrain runtime ${{ matrix.runtime }}
        uses: ./.git/nextstrain/.github/actions/setup-nextstrain-cli
        with:
          cli-version: ">=8.3.0"
          runtime: ${{ matrix.runtime }}
```

Unless there are objections, I'll refactor `run-nextstrain-build` to pass in the runtime as input.

actions/run-nextstrain-build-step/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{inputs.step }}-outputs-${{ matrix.runtime }}
+        path: |
+          ${{ inputs.step }}/results/
+          ${{ inputs.step }}/benchmarks/
+          ${{ inputs.step }}/logs/
+          ${{ inputs.step }}/.snakemake/log/
```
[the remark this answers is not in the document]
genehack: Nope, oversight; will re-add. (great catch, btw, thank you!)

(TL;DR: at point point I had three in-line steps instead of three calls to one composite action. the ingest build doesn't generate an auspice output, so the step for that didn't have it in the set of upload paths. when I refactored out the composition action, it was from the code doing the ingest step.)

actions/run-nextstrain-build-step/action.yaml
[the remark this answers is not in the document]
genehack: `actions/upload` has an option, `if-no-files-found:`; i'll set that to `ignore` and see if that makes this sufficiently un-annoying.

i had the upload as "always" with the thought that if the build failed partially, the logs etc. might still be useful. the other way to support that would be to set some sort of flag inside the `if all required files are here` branch. open to options here.

.github/workflows/pathogen-repo-ci.yaml
genehack: Changed to `Checkout ${{ inputs.repo }}`

.github/workflows/pathogen-repo-ci.yaml
genehack: fixed.

.github/workflows/pathogen-repo-ci.yaml
genehack: fixed.

actions/run-nextstrain-build/action.yaml
genehack: s/`STEP`/`DIR`/g

genehack:
> This repo's CI (`.github/workflows/ci.yaml`) is failing to actually test `pathogen-repo-ci.yaml` because it points at nextstrain/ebola and nextstrain/zika-tutorial, both of which result in all builds skipped.

I think the `zika-tutuorial` one can stay, because that's called `test-pathogen-repo-ci-no-example-data`, so everything being skipped seems to be the point under test. 

I changed `ebola` back to `zika` and removed the `TODO` comment.

.github/workflows/pathogen-repo-ci.yaml
genehack: The point of the CI workflow is supposed to be highly opinionated, low configuration, using a standard interface (the defined sub-dirs and the presence of a CI-specific build configuration).

Currently the output artifact is named: `name: ${{ inputs.directory }}-outputs-${{ inputns.runtime }}` — I think if I change that to something like `name: ci-outputs-${{ inputs.directory }}-${{ inputs.runtime }}` that aligns with the overall "no knobs" design.

.github/workflows/pathogen-repo-ci.yaml
genehack: but also, sans further commentary: 

```
> ag 'setup '
workflows/pathogen-repo-build.yaml.in
282:        name: Setup runtime ${{ inputs.runtime }}

workflows/pathogen-repo-build.yaml
308:      - name: Setup runtime ${{ inputs.runtime }}
385:      - name: Setup runtime ${{ inputs.runtime }}
443:      - name: Setup runtime ${{ inputs.runtime }}
502:      - name: Setup runtime ${{ inputs.runtime }}
561:      - name: Setup runtime ${{ inputs.runtime }}
643:      - name: Setup runtime ${{ inputs.runtime }}
```

actions/run-nextstrain-build/action.yaml
```diff
@@ -1,11 +1,11 @@
-name: run-nextstrain-build-step
+name: run-nextstrain-build
```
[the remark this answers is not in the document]
genehack: :+1:

actions/run-nextstrain-build-step/action.yaml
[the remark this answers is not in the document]
joverlee521: > the other way to support that would be to set some sort of flag inside the if all required files are here branch.

I like this route, just a flag of did we try to run `nextstrain build` seems enough. Could be helpful to see when the expected files aren't available?

genehack:
Okay, I think that I've addressed everything; ready for re-review.

tsibley:
> > 
> 
> I think the `zika-tutuorial` one can stay, because that's called `test-pathogen-repo-ci-no-example-data`, so everything being skipped seems to be the point under test.

Not quite. The point was to test that the (phylo) build ran even when example data was omitted (because it's optional before this PR). But because `zika-tutorial` doesn't follow the modern repo layout (i.e. `phylogenetic/`) nothing gets run at all.

genehack:
> I changed `ebola` back to `zika` and removed the `TODO` comment.

...and still nothing ran in `zika` because `zika` _isn't_ actually fully compliant with the repo guide. Will update to `seasonal-cov`, which I know is.

actions/run-nextstrain-build/action.yaml
```diff
@@ -1,14 +1,24 @@
...
 
 inputs:
   directory:
     description: The name of the sub-directory to run the build from
     type: string
     required: true
+  runtime:
+    description: Nextstrain runtime to use for the build
+    type: string
+    required: true
```
joverlee521: I think it's safer to verify and set the default runtime to match this runtime input so that we don't accidentally switch runtimes. 

In the `run-build` step, this can be done with 
```
nextstrain check-setup ${{ inputs.runtime }} --set-default
```

genehack: Added in 2687ba0

.github/workflows/pathogen-repo-ci.yaml
tsibley: Yeah, that's only one actual instance (in the `.in` file), but it's also wrong. (Not my fault though ;-)

.github/workflows/pathogen-repo-ci.yaml
tsibley: Not quite. The reason for the `artifact-name` input is so that this repo's own CI distinguishes artifacts between multiple invocations of `pathogen-repo-ci.yaml`.

genehack:
> > I changed `ebola` back to `zika` and removed the `TODO` comment.
> 
> ...and still nothing ran in `zika` because `zika` _isn't_ actually fully compliant with the repo guide. Will update to `seasonal-cov`, which I know is.
> 
> ETA: actually, it's not yet because that work is unmerged and blocks on _this_ work. _sigh_ Gonna add `nextstrain-pathogen.yaml` to `zika` i guess...

...and it still didn't run because the config file isn't named as expected _sigh_

.github/workflows/pathogen-repo-ci.yaml
genehack: At this point I am not sure what change is being requested. 

It sounds like you're wanting a new `archive-name` input for `pathogen-repo-ci` that would need to get passed down into `run-nextstrain-ci-build` — is that correct?

genehack:
> > > I changed `ebola` back to `zika` and removed the `TODO` comment.
> > 
> > 
> > ...and still nothing ran in `zika` because `zika` _isn't_ actually fully compliant with the repo guide. Will update to `seasonal-cov`, which I know is.
> > ETA: actually, it's not yet because that work is unmerged and blocks on _this_ work. _sigh_ Gonna add `nextstrain-pathogen.yaml` to `zika` i guess...
> 
> ...and it still didn't run because the config file isn't named as expected _sigh_

okay, [zika phylogenetic build runs in this repo's CI now](https://github.com/nextstrain/.github/actions/runs/9388770411/job/25855492291?pr=90).

.github/workflows/pathogen-repo-ci.yaml
genehack: @tsibley ☝️

tsibley: There are two invocations of `pathogen-repo-ci.yaml` in this repo's own `ci.yaml`. What I want is for the artifacts produced by those invocations to not conflict/merge with each other.

actions/run-nextstrain-ci-build/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+      run: |
+        if [[ -f nextstrain-pathogen.yaml && -f $DIR/Snakefile && -f $DIR/build-configs/ci/config.yaml ]]; then
+          nextstrain check-setup ${{ inputs.runtime }} --set-default
+          nextstrain build $DIR --configfile build-configs/ci/config.yaml
+        else
+          echo "Skipping $DIR build due to one or more missing files."
+          for i in nextstrain-pathogen.yaml $DIR/Snakefile $DIR/build-configs/ci/config.yaml; do
+            ! test -f $i && echo missing $i
+          done
+        fi
```
tsibley, line 38: _non-blocking_

Our current shellcheck action is missing this (I'll have a fix for that later), but some minor linting things here:

```suggestion
      env:
        DIR: ${{ inputs.directory }}
      run: |
        if [[ -f nextstrain-pathogen.yaml && -f "$DIR"/Snakefile && -f "$DIR"/build-configs/ci/config.yaml ]]; then
          nextstrain check-setup ${{ inputs.runtime }} --set-default
          nextstrain build "$DIR" --configfile build-configs/ci/config.yaml
        else
          echo "Skipping $DIR build due to one or more missing files."
          for i in nextstrain-pathogen.yaml "$DIR"/Snakefile "$DIR"/build-configs/ci/config.yaml; do
            ! test -f "$i" && echo missing "$i"
          done
        fi
```

I'd also gently suggest using `[[ ]]` instead of `test` for Reasons, e.g.

```bash
[[ -f "$i" ]] || echo missing "$i"
```

actions/run-nextstrain-ci-build/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+runs:
+  using: "composite"
+  steps:
+    - id: run-build
+      env:
+        DIR: ${{ inputs.directory }}
+      run: |
+        if [[ -f nextstrain-pathogen.yaml && -f $DIR/Snakefile && -f $DIR/build-configs/ci/config.yaml ]]; then
+          nextstrain check-setup ${{ inputs.runtime }} --set-default
+          nextstrain build $DIR --configfile build-configs/ci/config.yaml
```
victorlin, line 32: Echoing @joverlee521 from [Slack](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1716316838146949?thread_ts=1716315967.004439&cid=C0K3GS3J8): did you consider using a Snakemake target instead of a config file? This would looks something like:

```suggestion
            nextstrain build $DIR ci_target
```

config/defaults.yaml:

```yaml
viruses:
  - "229e"
  - "nl63"
  - "oc43"
  - "hku1"

ci_viruses:
  - "hku1"
```

Snakefile:

```Snakefile
rule ci_target:
    input:
        expand("auspice/{virus}.json", virus=config["ci_viruses"]),
```

actions/run-nextstrain-ci-build/action.yaml
tsibley: ShellCheck linting fix is https://github.com/nextstrain/.github/pull/91

actions/run-nextstrain-ci-build/action.yaml
joverlee521: _Debating with my past self here_

The Snakemake target only works if we are pointing to a target from the default workflow, it would not support using custom rules to copy over example data...

joverlee521: See [example in the pathogen-repo-guide](https://github.com/nextstrain/pathogen-repo-guide/blob/80cc1733b47f173772fe45005b82a69c6f8cb6c9/phylogenetic/build-configs/ci/config.yaml)

genehack: So, to wrap this up: no, I did not consider a Snakemake target; mainly because I was following what I saw in the pathogen-repo-guide. Gonna leave this as-is.

actions/run-nextstrain-ci-build/action.yaml
genehack: Fixed in 46abc25

.github/workflows/pathogen-repo-ci.yaml
genehack: Added in 3b49692

tsibley (approved):
Good to merge, I think.

FWIW, the commits on this branch are of the "all the messy details and false starts" style instead of "tells a story" style. I have a strong preference for the latter, but I won't try to insist on it.

genehack:
> FWIW, the commits on this branch are of the "all the messy details and false starts" style instead of "tells a story" style. 

To me, there is a story here — the story is, "here's what I originally thought would satisfy the rquirements, and here's all the stuff people told me in code review that I should change". 

If you have specific things to point to and say (e.g.), "instead of adding this commit, I would have squashed it back here" or "I would combine all the PR feedback into a single commit at the end", or something like that, I'm happy to listen to that feedback and try to take it into account in the future — but I assure you, I _am_ trying to tell a story with my branches, although perhaps more of a "lab notebook" version than a "journal article" recounting.

tsibley:
Got it. I didn't give specific feedback here because I didn't intend to insist on it. I'll try to give more specific feedback in the future.

Issue #92, opened 2024-06-10
Should the pathogen-repo-ci be louder if no workflows are run? 


joverlee521:
### TODO

- [x] move `nextstrain-pathogen.yaml` file checks out into a single step prior to workflow run steps
- [x] add aggregate success step, modeled on [`pathogen-repo-build` `wait-conclusion` logic](https://github.com/nextstrain/.github/blob/7f910e6c7d3d475500b063e4861a26b63c168cb8/.github/workflows/pathogen-repo-build.yaml#L596-L613) to trigger failure when no build steps are run

### Original issue content

Follow up to #89 

Just noticed in our [docker-base CI workflow](https://github.com/nextstrain/docker-base/actions/runs/9440522488) that all the pathogen-repo-ci jobs have "succeeded" even though none of them ran any workflows. 

This outcome is pretty misleading and can potentially lead to us missing when changes actually break pathogen workflows if we don't notice that the workflows never actually ran within CI.

joverlee521:
I think fine to leave each ingest/phylogenetic/nextclade workflow as optional, but maybe we should have a final conclusion job that checks at least _one_ of the workflows ran?

genehack:
> I think fine to leave each ingest/phylogenetic/nextclade workflow as optional, but maybe we should have a final conclusion job that checks at least _one_ of the workflows ran?

In this _particular_ case, every repo where the `pathogen-repo-ci` workflow is being used lacks the `nextstrain-pathogen.yaml` file in the repo root. 

Maybe instead of testing for the presence of that file in the individual build steps, and skipping if it's absent, there should be a preliminary job that checks once, and generates a failure (e.g., `exit 1` or the like) if the file is missing.

joverlee521:
> In this _particular_ case, every repo where the `pathogen-repo-ci` workflow is being used lacks the `nextstrain-pathogen.yaml` file in the repo root.
> 
> Maybe instead of testing for the presence of that file in the individual build steps, and skipping if it's absent, there should be a preliminary job that checks once, and generates a failure (e.g., `exit 1` or the like) if the file is missing.

Yup, definitely support erroring early on this! In addition, I still think it'd be helpful to fail when _none_ of the expected workflows ran.

genehack:
> Yup, definitely support erroring early on this! In addition, I still think it'd be helpful to fail when _none_ of the expected workflows ran.

You wouldn't happen to be able to point me to an example of that sort of aggregated failure check in a GitHub Action, would you?

[...]

PR #95, opened 2024-06-11
Make `pathogen-repo-ci` fail when config is missing or no builds are attempted

genehack:

Make the CI fail if it doesn't try to run at least one build step, as that's an indication that somebody is trying to use "modern" `pathogen-repo-ci` on an un-modernized pathogen repo (or there's some other grievous misconfig happening).  

## Related issue(s)

#92 

## Checklist

- [x] Clean up CI bits in this repo that break
- [x] Checks pass
- [x] Verify a repo without `pathogen-nextstrain.yaml` fails as expected [proof](https://github.com/nextstrain/seasonal-cov/actions/runs/9469661700/job/26089377134#step:8:12)
- [x] Verify a repo that has `pathogen-nexstrain.yaml` but lacking required files for all build steps fails as expected [proof](https://github.com/nextstrain/seasonal-cov/actions/runs/9469995788/job/26089927906)
- [x] Verify a repo that has one step that attempts to run succeeds as expected [proof](https://github.com/nextstrain/seasonal-cov/actions/runs/9470059750/job/26090113771)
- [x] Verify a repo that runs two steps still succeeds as expected [proof](https://github.com/nextstrain/seasonal-cov/actions/runs/9470094934/job/26090223884)
- [ ] ~Verify a repo that runs all three steps still succeeds as expected~ — skipping because I don't think there's a repo handy with a `nextclade` build that's also using pathogen-repo-ci(?)

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -238,6 +238,13 @@ jobs:
           ref: ${{ needs.workflow-context.outputs.sha }}
           path: ${{ env.NEXTSTRAIN_GITHUB_DIR }}
 
+      - name: Verify nextstrain-pathogen.yaml file
+        run: >
+          if [[ ! -f nextstrain-pathogen.yaml ]]; then
+            echo "To use this workflow, there must be a 'nextstrain-pathogen.yaml' file present in the repository root";
+            exit 1;
+          fi
+
       - name: Set up Nextstrain runtime ${{ matrix.runtime }}
         uses: ./.git/nextstrain/.github/actions/setup-nextstrain-cli
         with:
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -252,6 +252,7 @@ jobs:
           runtime: ${{ matrix.runtime }}
 
       - name: Run ingest
+        id: ingest
         uses: ./.git/nextstrain/.github/actions/run-nextstrain-ci-build
         with:
           directory: ingest
@@ -259,6 +260,7 @@ jobs:
           artifact-name: ${{ inputs.artifact-name }}
 
       - name: Run phylogenetic
+        id: phylogenetic
         uses: ./.git/nextstrain/.github/actions/run-nextstrain-ci-build
         with:
           directory: phylogenetic
@@ -266,8 +268,19 @@ jobs:
           artifact-name: ${{ inputs.artifact-name }}
 
       - name: Run nextclade
+        id: nextclade
         uses: ./.git/nextstrain/.github/actions/run-nextstrain-ci-build
         with:
           directory: nextclade
           runtime: ${{ matrix.runtime }}
           artifact-name: ${{ inputs.artifact-name }}
+
+      - name: Verify a workflow ran
+        run: |
+          # shellcheck disable=SC2129,SC2242
+
+          # if we see at least one success, we're good
+          echo "INGEST ATTEMPTED=${{ steps.ingest.outputs.run-attempted }}" >> "$GITHUB_STEP_SUMMARY"
+          echo "PHYLOGENETIC ATTEMPTED=${{ steps.phylogenetic.outputs.run-attempted }}" >> "$GITHUB_STEP_SUMMARY"
+          echo "NEXTCLADE ATTEMPTED=${{ steps.nextclade.outputs.run-attempted }}" >> "$GITHUB_STEP_SUMMARY"
+          exit ${{ contains(steps.*.outputs.run-attempted, 'true') && '0' || '1' }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,12 +30,6 @@ jobs:
       artifact-name: outputs-test-pathogen-repo-ci
     secrets: inherit
 
-  test-pathogen-repo-ci-no-example-data:
-    uses: ./.github/workflows/pathogen-repo-ci.yaml
-    with:
-      repo: nextstrain/zika-tutorial
-      artifact-name: outputs-test-pathogen-repo-ci-no-example-data
-
   test-docs-ci-conda:
     uses: ./.github/workflows/docs-ci.yaml
     with:
--- a/actions/run-nextstrain-ci-build/action.yaml
+++ b/actions/run-nextstrain-ci-build/action.yaml
@@ -33,12 +33,12 @@ runs:
       env:
         DIR: ${{ inputs.directory }}
       run: |
-        if [[ -f nextstrain-pathogen.yaml && -f "$DIR"/Snakefile && -f "$DIR"/build-configs/ci/config.yaml ]]; then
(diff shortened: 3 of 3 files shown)


genehack:
The one remaining thing here is figuring out what to do with `CI / test-pathogen-repo-ci-no-example-data`, which is pointed at `zika-tutorial`. Specifically, this job:

```yaml
  test-pathogen-repo-ci-no-example-data:
    uses: ./.github/workflows/pathogen-repo-ci.yaml
    with:
      repo: nextstrain/zika-tutorial
      artifact-name: outputs-test-pathogen-repo-ci-no-example-data
```

options:

1. pin this job to the older workflow — feels bad, not sure what that proves
2. remove this job — would need somebody to confirm that's okay; i don't understand what this job is trying to verify from a CI perspective
3. point the job at a different repo — again, requires an explanation of the intent of this job in the CI workflow for the repo

@joverlee521 @tsibley could either of you shed insight on the "what is this job _for_" question?

tsibley:
> @joverlee521 @tsibley could either of you shed insight on the "what is this job _for_" question?

Ah, I [tried to answer that previously](https://github.com/nextstrain/.github/pull/90#issuecomment-2148363986), but sorry there's still a disconnect here. In retrospect, I think I assumed too much context. I think the commit which introduces the job, 9800e33cac40329e55357c584097c4b4249492a5, is a clear and concise explanation with the right context, but if it's still not I can try again.

genehack:
> > @joverlee521 @tsibley could either of you shed insight on the "what is this job _for_" question?
> 
> Ah, I [tried to answer that previously](https://github.com/nextstrain/.github/pull/90#issuecomment-2148363986), but sorry there's still a disconnect here. In retrospect, I think I assumed too much context. I think the commit which introduces the job, [9800e33](https://github.com/nextstrain/.github/commit/9800e33cac40329e55357c584097c4b4249492a5), is a clear and concise explanation with the right context, but if it's still not I can try again.

So, based on the context in that commit message, I think the right thing to do here is just remove this job from the CI? The thing that it's validating (don't need to set up if example data doesn't exist) is something that the "modern" `pathogen-repo-ci` doesn`t worry about. The assumption is if there's a Snakefile and a `build-config/ci/config.yaml`, it's sufficient to call the build with that config and it will DTRT.

genehack:
This is updated and seems to be working; will try to merge this EOD Wednesday (PT), if not earlier.

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -238,6 +238,16 @@ jobs:
           ref: ${{ needs.workflow-context.outputs.sha }}
           path: ${{ env.NEXTSTRAIN_GITHUB_DIR }}
 
+      - name: Verify nextstrain-pathogen.yaml file
+        continue-on-error: false
```
tsibley, line 242: _nit_

This is the default, and we don't make it explicit elsewhere. I'd drop the line.

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -238,6 +238,16 @@ jobs:
...
+        run: >
+          # shellcheck disable=SC2154
```
tsibley, line 244: I don't think disabling this check is necessary.

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -238,6 +238,16 @@ jobs:
...
+
+          [[ -f nextstrain-pathogen.yaml ]] || (
+               echo "To use this workflow, there must be a 'nextstrain-pathogen.yaml' file present in the repository root" \
+            && exit 1
+          )
```
tsibley, line 249: _nit_

A more standard construction of this would be to use an `if` statement…

```suggestion
          if [[ ! -f nextstrain-pathogen.yaml ]]; then
               echo "To use this workflow, there must be a 'nextstrain-pathogen.yaml' file present in the repository root"
               exit 1
          fi
```

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -255,22 +255,36 @@ jobs:
...
       - name: Run nextclade
+        id: nextclade
         uses: ./.git/nextstrain/.github/actions/run-nextstrain-ci-build
         with:
           directory: nextclade
           runtime: ${{ matrix.runtime }}
           artifact-name: ${{ inputs.artifact-name }}
+
+      - name: Verify a workflow ran
+        continue-on-error: false
```
tsibley, line 282: _nit_

Ditto re: `continue-on-error: false` being the default we rely on everywhere.

genehack:
> My biggest suggestion is that we don't actually run the "Run {ingest, phylogenetic, nextclade}" steps if they're not going to do anything. 

The advantage I see to doing the file existence checks inside the action is that it can report out exactly which file(s) aren't present. Maybe you can do that with some sort of `if:` expression — my GitHub Actions fu remains weak — and that would be fine, but having it all sort of collected in the one spot makes it easier (at least for me) to understand.

tsibley (approved):

tsibley:
Take a gander at https://github.com/nextstrain/.github/pull/96 and see what you think?

> The advantage I see to doing the file existence checks inside the action is that it can report out exactly which file(s) aren't present.

Nod. Reporting that out is still possible, but I didn't bother in the PR above. It seems like a minor thing to me: if a step is skipped unexpectedly, it's easy enough to grab the commit id checked out from the logs and look at what files exist in it? But we could still maintain that reporting if desired.

.github/workflows/pathogen-repo-ci.yaml
```diff
@@ -238,29 +238,49 @@ jobs:
...
           artifact-name: ${{ inputs.artifact-name }}
+
+      - name: Verify a workflow ran
+        run: |
+          # shellcheck disable=SC2242
+
+          # if we see at least one success, we're good
+          echo "INGEST ATTEMPTED=${{ steps.ingest.outputs.run-attempted }}"
+          echo "PHYLOGENETIC ATTEMPTED=${{ steps.phylogenetic.outputs.run-attempted }}"
+          echo "NEXTCLADE ATTEMPTED=${{ steps.nextclade.outputs.run-attempted }}"
```
joverlee521, line 285: Minor suggestion to output these to the `$GITHUB_STEP_SUMMARY` so that it's easy to scan the workflow attempts from the GH Action summary instead of having to dig into the job logs.

joverlee521 (approved):
Thanks for the work here @genehack! I left a minor suggestion, but the changes here LGTM!

.github/workflows/pathogen-repo-ci.yaml
genehack: pushed an update — was that what you were thinking of?

joverlee521: Yup, that's what I was thinking, except small typo `$"GITHUB_STEP_SUMMARY"` -> `"$GITHUB_STEP_SUMMARY"`

genehack: >_< thanks -- pushed a fix, gonna merge this once CI finishes

joverlee521: Screenshot 2024-06-13 at 10 18 08 AM

🎉

PR #98, opened 2024-06-14
pathogen-repo-ci: Keep running even if one of the pathogen workflows fails

tsibley:
In testing it's helpful to report all test results even in the face of an early failure to reduce the chances of playing whack-a-mole with unreported failures once when you fix the first failure.  For example, we want to know if the phylogenetic workflow fails even when the ingest workflow run before it fails.

## Checklist

- [x] Checks pass
- [x] Test that failure behaviour works as expected ([no valid workflows](https://github.com/nextstrain/.github/actions/runs/9511668562), [two valid workflows](https://github.com/nextstrain/.github/actions/runs/9511672955), [one broken workflow](https://github.com/nextstrain/.github/actions/runs/9523126166))

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -286,16 +286,18 @@ jobs:
         if: hashFiles('ingest/Snakefile') && hashFiles('ingest/build-configs/ci/config.yaml')
         id: ingest
         run: nextstrain build ingest --configfile build-configs/ci/config.yaml
+        continue-on-error: true
       - name: Run phylogenetic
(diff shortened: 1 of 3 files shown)


joverlee521 (approved):
Seems reasonable to me 👍

tsibley:
Hmm, `continue-on-error: true` is maybe not quite how we want to achieve this behaviour, though. It results in the right behaviour, but misleading job step indicators: both of these failed but aren't marked as such in this bit of UI.

image


I wish I'd realized this a minute earlier. I think I can swap around implementation to have our cake and eat it too.

PR #99, opened 2024-06-14
pathogen-repo-ci: Use status functions in conditionals instead of continue-on-error: true

tsibley:
While "continue-on-error: true" resulted in the behaviour we wanted, it also meant the job step status indicators were misleading: even when an ingest/phylogenetic/nextclade step failed, the UI showed a check mark for them in the logs.  Instead of masking failures to keep going in the face of errors, override the default success() conditional by specifying our own conditions on a status expression.

Related-to: <https://github.com/nextstrain/.github/pull/98#issuecomment-2168846812>

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/pathogen-repo-ci.yaml
+++ b/.github/workflows/pathogen-repo-ci.yaml
@@ -286,18 +286,16 @@ jobs:
         if: hashFiles('ingest/Snakefile') && hashFiles('ingest/build-configs/ci/config.yaml')
         id: ingest
         run: nextstrain build ingest --configfile build-configs/ci/config.yaml
-        continue-on-error: true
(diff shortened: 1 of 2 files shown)


tsibley:
[Much better](https://github.com/nextstrain/.github/actions/runs/9523269073/job/26254387097):

image


and still works with [no valid workflows](https://github.com/nextstrain/.github/actions/runs/9523262191/job/26254368911) and [two valid workflows](https://github.com/nextstrain/.github/actions/runs/9523265809/job/26254377135). I wish we had testing of that in CI for this repo so I didn't have to keep manually triggering them, but that cracks the lid on another kettle of fish, so I won't for now.

PR #104, opened 2024-08-05
Add `make linkcheck` step, to check external links [#29]

### LINKED ISSUE #29: docs-ci: Check external links
Running Sphinx's `linkcheck` builder to systematically check our external links would be really helpful to catch broken links resulting from anything from typos when authoring to third-party site reorganizations.

It would have caught a bunch of external link errors in https://github.com/nextstrain/docs.nextstrain.org/pull/139, for example. (Not calling out that PR specifically! This sort of error is very common and easy to make. But that PR made me wonder why we weren't already running `linkcheck`.)

A second job (for parallelism) or just a subsequent step in the current job should run something like:

    SPHINXOPTS="-W --keep-going" make linkcheck

which will produce a bunch of messages to stdout/stderr output but also `build/linkcheck/output.{txt,json}`. Both of the latter files (and probably just the whole `build/` dir!) should be uploaded as a workflow artifact. [Examples of output](https://gist.github.com/tsibley/40f3a57a2e68f578a7dff51692bc9cc3) from https://github.com/nextstrain/docs.nextstrain.org/commit/6fcf76f.

Since `linkcheck` will exit with a non-zero status on errors, it'll fail the CI job. I believe this is what we want.

[Some configuration](https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder) for `linkcheck` is usually required to ignore links that are actually ok, and this would fall on each doc project to configure itself.

## related issues:
- [x] nextstrain/docs.nextstrain.org#14

genehack:

Adds a `make linkcheck` step to the Sphinx CI.

Note that this blocks on the various PRs attached to nextstrain/docs.nextstrain.org#14 — those PRs should be merged before this one.

Update to add: I realized this workflow is used by more than `docs.nextstrain.org`, so before this can be merged, the following repos need to be audited for broken links:

- [x] Audit augur -- nextstrain/augur#1570
- [x] Audit auspice --  nextstrain/auspice#1808
- [x] Audit cli -- nextstrain/cli#389
- [x] Audit ncov -- nextstrain/ncov#1137

## Related issue(s)

Closes #29 
nextstrain/docs.nextstrain.org#14

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/docs-ci.yaml
+++ b/.github/workflows/docs-ci.yaml
@@ -68,6 +68,9 @@ jobs:
       - run: make ${{ inputs.make-target }}
         working-directory: ${{ inputs.docs-directory }}
 
+      - run: make linkcheck
+        working-directory: ${{ inputs.docs-directory }}
+
   build-pip:
     if: inputs.pip-install-target != ''
     runs-on: ubuntu-latest
@@ -81,4 +84,7 @@ jobs:
       - run: pip list
 
       - run: make ${{ inputs.make-target }}
+        working-directory: ${{ inputs.docs-directory }}
+
+      - run: make linkcheck
         working-directory: ${{ inputs.docs-directory }}


victorlin (approved):

.github/workflows/docs-ci.yaml
```diff
@@ -82,3 +82,6 @@ jobs:
 
       - run: make ${{ inputs.make-target }}
         working-directory: ${{ inputs.docs-directory }}
+
+      - run: SPHINXOPTS="-W --keep-going" make linkcheck ${{ inputs.make-target }}
```
joverlee521, line 86: _non-blocking question_ 

Is `SPHINXOPTS="-W --keep-going"` meant to override the default envvar set in the workflow? 
https://github.com/nextstrain/.github/blob/429f042333c3cbfd2587d2942d283a57d3d252cc/.github/workflows/docs-ci.yaml#L48

genehack: > _non-blocking question_
> 
> Is `SPHINXOPTS="-W --keep-going"` meant to override the default envvar set in the workflow?
> 
> https://github.com/nextstrain/.github/blob/429f042333c3cbfd2587d2942d283a57d3d252cc/.github/workflows/docs-ci.yaml#L48

TBPH, I copied that invocation [from Tom's original note in the issue](https://github.com/nextstrain/.github/issues/29#issue-1442955952), and I didn't notice there was a default setting. Experimenting locally, `-n` doesn't seem to make a difference one way or the other, so I'll remove the env var in this step.

joverlee521 (changes requested):
I think we'd want to add the new `make linkcheck` step the the `build-conda` job as well?

genehack:
> I think we'd want to add the new `make linkcheck` step the the `build-conda` job as well?

Good call, done.

Also removed env-var and adjusted use of `make-target`.

joverlee521 (approved):

genehack:
Updated description above to note that before this can land, everything that uses it (listed above) needs to be manually audited to get `make linkcheck` to run cleanly. Links to open issues above.

.github/workflows/docs-ci.yaml
```diff
@@ -82,3 +85,6 @@ jobs:
 
       - run: make ${{ inputs.make-target }}
         working-directory: ${{ inputs.docs-directory }}
+
+      - run: make linkcheck
+        working-directory: ${{ inputs.docs-directory }}
```
victorlin, line 90: > I realized this workflow is used by more than `docs.nextstrain.org`, so before this can be merged, the following repos need to be audited for broken links:

An alternative is to make this opt-in. This can be achieved by specifying a new workflow input (e.g. `linkcheck-target` and conditioning this step on its presence:

```
- if: inputs.linkcheck-target
  run: make ${{ linkcheck-target }}
  working-directory: ${{ inputs.docs-directory }}
```

Usage would look like:

```
  docs-ci:
    uses: nextstrain/.github/.github/workflows/docs-ci.yaml@master
    with:
      environment-file: conda.yml
      docs-directory: .
      make-target: html
      linkcheck-target: linkcheck
```

genehack: I've got momentum on this now, I'll just keep rolling through the other two repos...

genehack:
This is now ready to merge — all the affected downstream repos should have clean `make linkcheck` runs in their doc trees — but I'm going to wait until Monday to actually push the button, just in case something goes pear-shaped after all.

genehack:
> This is now ready to merge — all the affected downstream repos should have clean `make linkcheck` runs in their doc trees — but I'm going to wait until Monday to actually push the button, just in case something goes pear-shaped after all.

I also validated a branch in each of the five repos that consume this PI action, changing `docs-ci.yaml` to point to this branch, and verifying that the CI remained green.

PR #107, opened 2024-09-03
Split `make linkcheck` out into a distinct job [#106]

### LINKED ISSUE #106: docs-ci: `make linkcheck` prone to transient network failures
I've just run into [this error](https://github.com/nextstrain/augur/actions/runs/10530367718/job/29180176987#step:6:731) on an Augur PR which did not change any docs links:

```
(api/developer/augur.merge: line    7) broken    https://www.gnu.org/software/bash/manual/bash.html#ANSI_002dC-Quoting - HTTPSConnectionPool(host='www.gnu.org', port=443): Max retries exceeded with url: /software/bash/manual/bash.html (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f4835cbc1c0>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
build finished with problems.
make: *** [Makefile:20: linkcheck] Error 1
```

This seems like a transient network error which shows up as a failing check ❌ on the PR which confused me at first.

## Possible solutions

1. ⛔️ Split `linkcheck` into a separate job on docs-ci and use `continue-on-error: true`
    - https://github.com/nextstrain/.github/pull/107
    - Not a great solution because it masks real linkcheck issues
3. Don't run `linkcheck` in CI but instead on a weekly schedule
4. Ignore `timeout` codes
    - https://github.com/nextstrain/.github/pull/110
    - This should be a good improvement but does not cover all transient failures (see below)
5. (per-project) Handle `broken` codes on a case-by-case basis. If it's intermittent, add to `linkcheck_ignore`. Otherwise, update or remove the link.

genehack:

Split `make linkcheck` into a distinct job, with `continue-on-error: true`, so that transient failures don't show as CI fails. 
## Related issue(s)

Closes #106 

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/docs-ci.yaml
+++ b/.github/workflows/docs-ci.yaml
@@ -68,9 +68,6 @@ jobs:
       - run: make ${{ inputs.make-target }}
         working-directory: ${{ inputs.docs-directory }}
 
-      - run: make linkcheck
-        working-directory: ${{ inputs.docs-directory }}
-
   build-pip:
     if: inputs.pip-install-target != ''
     runs-on: ubuntu-latest
@@ -86,5 +83,17 @@ jobs:
       - run: make ${{ inputs.make-target }}
         working-directory: ${{ inputs.docs-directory }}
 
-      - run: make linkcheck
+  linkcheck:
+    if: inputs.pip-install-target != ''
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.repo }}
+
+      - run: pip install '${{ inputs.pip-install-target }}'
+
+      - id: linkcheck
+        run: make linkcheck
         working-directory: ${{ inputs.docs-directory }}
+        continue-on-error: true


.github/workflows/docs-ci.yaml
```diff
@@ -86,5 +85,17 @@ jobs:
       - run: make ${{ inputs.make-target }}
         working-directory: ${{ inputs.docs-directory }}
 
-      - run: make linkcheck
+  linkcheck:
+    if: inputs.pip-install-target != ''
```
victorlin, line 89: This prevents linkcheck from running on conda-based build environments. It should be split into `linkcheck-conda` and `linkcheck-pip`.

Another approach would be what @tsibley proposed in https://github.com/nextstrain/.github/pull/28#pullrequestreview-1173000933. This would allow for a single job to handle both conda and pip, but it's more complex:

> My idea had been to make this workflow read the RTD config file and install what's configured there, which I still think would be nicer, but it's more complex.

.github/workflows/docs-ci.yaml
```diff
@@ -39,6 +39,8 @@ on:
         type: string
         default: html
 
+permissions:
+  pull-requests: write
```
victorlin, line 43: PR description says `continue-on-error-comment` is added but I don't see it here, just this new permissions block which seems related. I assume this is pending [feedback on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1725402169663899)?

victorlin (changes requested):

.github/workflows/docs-ci.yaml
genehack: > This prevents linkcheck from running on conda-based build environments. It should be split into `linkcheck-conda` and `linkcheck-pip`.

I don't see the utility of running the exact same linkcheck process twice — that's just going to cause additional fails due to random network glitches.

In what scenario does a link resolve in the conda build and fail to resolve in the pip build, or vice versa?

.github/workflows/docs-ci.yaml
genehack: Missed a couple bits when I ripped `continue-on-error-comment` back out -- fixed.

.github/workflows/docs-ci.yaml
victorlin: `build-conda` and `build-pip` are meant to be mutually exclusive. This is not checked in the workflow because it's reasonable to assume that a docs project will use only one. See GitHub query [`org:nextstrain nextstrain/.github/.github/workflows/docs-ci.yaml`](https://github.com/search?q=org%3Anextstrain+nextstrain%2F.github%2F.github%2Fworkflows%2Fdocs-ci.yaml&type=code): some invocations specify `environment-file` while others specify `pip-install-target`.

victorlin:
Closing this in favor of https://github.com/nextstrain/.github/pull/110 as a solution to https://github.com/nextstrain/.github/issues/106

PR #109, opened 2024-10-11
Use a single build job

victorlin:

The number of steps that are shared between the two jobs now exceeds the number of steps that differ. In this case, it seems more maintainable to use one job with per-step conditions for environment setup.

This should also look nicer downstream - previously there was always one job that was skipped since they are meant to be mutually exclusive.

## Checklist

- [ ] Checks pass

### DIFF
--- a/.github/workflows/docs-ci.yaml
+++ b/.github/workflows/docs-ci.yaml
@@ -48,8 +48,7 @@ env:
   SPHINXOPTS: -n -W --keep-going
 
 jobs:
-  build-conda:
-    if: inputs.environment-file != ''
+  build:
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-ci.yaml
```diff
@@ -48,8 +48,7 @@ env:
   SPHINXOPTS: -n -W --keep-going
 
 jobs:
-  build-conda:
-    if: inputs.environment-file != ''
+  build:
```
genehack, line 51: Might be a nice safe-guard to have an early step that checks for both `inputs.environment-file` and `inputs.pip-install-target` being `== ''` and bails out with an error about the config being wrong. 

(I'm thinking of situations like somebody typo-ing in the config or something…)

genehack (approved):

.github/workflows/docs-ci.yaml
victorlin: Same for having both `environment-file` and `pip-install-target` for whatever reason. It'll probably work fine but we shouldn't have a use case for that.

I'll make a separate PR after merging this one.

victorlin: https://github.com/nextstrain/.github/pull/111

Issue #119, opened 2024-11-26
keep-workflows-enabled: query for scheduled GH Action workflows


joverlee521:
### Context
Original feedback from @tsibley in https://github.com/nextstrain/.github/pull/118#discussion_r1859001214

> I'd much prefer to query these (e.g. from GitHub's API) rather than hardcode them all here... (but we don't need to do that now if you aren't so inclined)

### Description

Replace the [hard-coded list of workflows](https://github.com/nextstrain/.github/blob/25cc821a4f94d1bdf55d1307ec8d22c6530a7edb/.github/workflows/keep-workflows-enabled.yaml#L23-L49) with API query to find all Nextstrain GH Action workflows that have a `schedule` event to automatically keep them enabled. This will lessen the burden of need to update the central list whenever we add a new scheduled workflow.

PR #118, opened 2024-11-25
Keep workflows enabled

joverlee521:
## Description of proposed changes

Uses the GH API to "enable" scheduled workflows on the first of
every month as a work-around for scheduled workflows being automatically
disabled when no repository activity has occurred in 60 days.

Included all active Nextstrain org GH Action workflows that have the
`schedule` trigger plus this new workflow so it keeps itself enabled.

<https://github.com/search?q=org%3Anextstrain+path%3A.github%2Fworkflows%2F**+schedule&type=code>

## Related issue(s)

Resolves https://github.com/nextstrain/.github/issues/112

## Checklist

- [x] Checks pass
- [x] [Test run](https://github.com/nextstrain/.github/actions/runs/12036559030)

PR #120, opened 2024-11-26
Keep yellow-fever workflows enabled too.

genehack:

I noticed yellow-fever wasn't in the recent #118; this adds it to the list. 

## Related issue(s)

#118 #119 

## Checklist

- [ ] Checks pass

### DIFF
--- a/.github/workflows/keep-workflows-enabled.yaml
+++ b/.github/workflows/keep-workflows-enabled.yaml
@@ -46,6 +46,7 @@ jobs:
           - { repo: seasonal-cov,    workflow: ingest-to-phylogenetic.yaml }
           - { repo: status,          workflow: ci.yaml }
           - { repo: WNV,             workflow: ingest-to-phylogenetic.yaml }
+          - { repo: yellow-fever,    workflow: ingest-to-phylogenetic.yaml }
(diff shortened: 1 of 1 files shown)


.github/workflows/keep-workflows-enabled.yaml
```diff
@@ -46,6 +46,8 @@ jobs:
           - { repo: seasonal-cov,    workflow: ingest-to-phylogenetic.yaml }
           - { repo: status,          workflow: ci.yaml }
           - { repo: WNV,             workflow: ingest-to-phylogenetic.yaml }
+          - { repo: yellow-fever,    workflow: ci.yaml }
```
joverlee521, line 49: The [yellow-fever CI](https://github.com/nextstrain/yellow-fever/blob/04b92ca46b3f24b11d47178c088e888dc4b4650b/.github/workflows/ci.yaml#L3-L8) does not have a `schedule` event so I don't think it needs to be included here.

joverlee521 (commented):
Depends on merge of https://github.com/nextstrain/yellow-fever/pull/20.

.github/workflows/keep-workflows-enabled.yaml
genehack: Ah, yeah, okay, removed the ci workflow.

PR #121, opened 2025-02-08
Add reusable workflow for workflow failure reporting

victorlin:

Taken from conda-base after generalizations¹ and further adjustments.

¹ <https://github.com/nextstrain/conda-base/compare/d7d23af57cf640124157e2c5c1c60a369b7931b1...cd33c38f0181d6451c8d8d086e24bc9e9c5364ae>

## Related issue(s)

Motivated by desire to do the same thing in another repository: https://github.com/nextstrain/cli/issues/412

## Checklist

- [ ] Checks pass
- [ ] @victorlin to add testing notes


.github/workflows/report-failure.yaml
```diff
@@ -0,0 +1,87 @@
...
+          echo "workflow_link=https://github.com/${{ github.repository }}/blob/${{ github.sha }}/$workflow_path" >> "$GITHUB_ENV"
+
+      - name: Create or close issue
+        run: |
+          # Search for issues opened by this job.
+          matching_issues=$(gh issue list \
+            --state open \
+            --author '@me' \
+            --json 'number,title' \
+            --jq 'map(select(.title == "${{ inputs.title }}") | .number)')
```
joverlee521, line 41: _non-blocking question_

Should there be some filter on the branch of the workflow? Or I guess the calling workflow can control if `report-failure.yaml` only gets run on the default branch...

joverlee521 (approved):
Thanks @victorlin! This should be useful for a lot of repos. We can even consider if we want to use it for the automated pathogen workflows.

I'm considering using this for the [nextstrain/status CI workflow](https://github.com/nextstrain/status/actions/workflows/ci.yaml). However, that workflow has a [history of temporary failures](https://github.com/nextstrain/status/actions/workflows/ci.yaml?query=is%3Afailure), so it might be get noisy.

victorlin:
> I'm considering using this for the [nextstrain/status CI workflow](https://github.com/nextstrain/status/actions/workflows/ci.yaml). However, that workflow has a [history of temporary failures](https://github.com/nextstrain/status/actions/workflows/ci.yaml?query=is%3Afailure), so it might be get noisy.

[Same thing](https://github.com/nextstrain/cli/issues/411#issuecomment-2644073326) with the nextstrain/cli CI workflow. I think in general it will be nice to have the issue automation in place, even if we decide to disable it temporarily during periods of unstable CI. If we find ourselves disabling it more than it's enabled, it's maybe a sign that CI status is not useful and that we should rework the steps to improve reliability.

.github/workflows/report-failure.yaml
victorlin: I think that's outside the scope of this reusable workflow. There could be a valid reason for the calling workflow to open issues regardless of branch. Yes, I've encoded this in the calling workflow for [conda-base](https://github.com/nextstrain/conda-base/blob/a4c8ddaa6b3b6c8c9e0b9706a208f0bb015025d8/.github/workflows/installation.yaml#L47), [cli](https://github.com/nextstrain/cli/blob/ad12f213a8bcae7b94a193956ea2dff6713a4013/.github/workflows/ci.yaml#L394) PRs.

joverlee521: Ah yeah that makes sense. I guess the title can include the branch of the workflow to make sure this matches issues per branch. 

Maybe the `inputs.title.description` should explicitly state that the title is used for issue matching.

victorlin: Done as part of cfa7538fb6ed08168983b8964f2299bdd8111a99

.github/workflows/report-failure.yaml
```diff
@@ -0,0 +1,87 @@
...
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract workflow link
+        run: |
+          workflow_path=$(echo "${{ github.workflow_ref }}" | sed -E 's|^.*/(\.github/workflows/[^@]+)@\S*$|\1|')
+          echo "workflow_link=https://github.com/${{ github.repository }}/blob/${{ github.sha }}/$workflow_path" >> "$GITHUB_ENV"
```
tsibley, line 40: I strongly suggest we avoid shell escaping issues (e.g. `"` in `github.workflow_ref`) by using environment variables instead of interpolations during YAML preprocessing. The default environment is enough; no need for step-specific ones to hold the context vars.

```suggestion
          workflow_path=$(echo "$GITHUB_WORKFLOW_REF" | sed -E 's|^.*/(\.github/workflows/[^@]+)@\S*$|\1|')
          echo "workflow_link=https://github.com/$GITHUB_REPOSITORY/blob/$GITHUB_SHA/$workflow_path" >> "$GITHUB_ENV"
```

Follow-on suggestion: Avoid sed and a regexp substitution and do simpler fixed prefix/suffix stripping.

```suggestion
          workflow_path="$GITHUB_WORKFLOW_REF"
          workflow_path="${workflow_path#"$GITHUB_REPOSITORY"/}"
          workflow_path="${workflow_path%%@*}"
          echo "workflow_link=https://github.com/$GITHUB_REPOSITORY/blob/$GITHUB_SHA/$workflow_path" >> "$GITHUB_ENV"
```

(Either way, note that workflow filenames can contain `@`, so none of the code will work in that case, but it won't be unsafe.)

.github/workflows/report-failure.yaml
tsibley, line 54: `inputs.title` (here and below) should be mapped to an environment variable (e.g. `env: TITLE: ${{ inputs.title }}`) and accessed as such in the jq code (e.g. `env.TITLE`). Otherwise quotes in the title cause havoc.

.github/workflows/report-failure.yaml
```diff
@@ -0,0 +1,87 @@
...
+            for issue in $(jq -r '.[]' <<<"$matching_issues"); do
+                echo "- https://github.com/${{ github.repository }}/issues/$issue" >&2
+            done
+            echo "This must be handled manually." >&2
+            exit 1
+          fi
+
+          existing_issue=$(jq -r '.[0] | values' <<<"$matching_issues")
+
+          # shellcheck disable=SC2050
```
tsibley, line 69: _non-blocking_

Oh, hmm. This is only necessary because of the way our Shellcheck wrapper for GitHub Actions works around the `${{ }}` interpolations. That's something we could probably improve. The point of that workaround to avoid having to tell Shellcheck to knock it off, but that's clearly not enough for this case.

.github/workflows/report-failure.yaml
```diff
@@ -0,0 +1,87 @@
...
+  workflow_call:
+    inputs:
+      failure:
+        description: Whether to create an issue indicating a failed workflow run, or keep an existing issue open.
+        type: boolean
+        required: true
+      success:
+        description: Whether to close any open issues created by this workflow.
+        type: boolean
+        required: true
```
tsibley, line 13: Taking both `failure` and `success` is weird, especially when the action taken on them (below) is intended to be mutually-exclusive (e.g. `assert failure xor success`). Why can't this be a single boolean?

tsibley (changes requested):

tsibley:
@joverlee521 

> I'm considering using this for the [nextstrain/status CI workflow](https://github.com/nextstrain/status/actions/workflows/ci.yaml). However, that workflow has a [history of temporary failures](https://github.com/nextstrain/status/actions/workflows/ci.yaml?query=is%3Afailure), so it might be get noisy.

Hmm. For that repo, I think the status page itself being stale should be enough to alert anyone who cares? This was part of my reason for [adding](https://github.com/nextstrain/status/pull/38) the staleness warning.

victorlin:
Noting that this + the other 2 PRs got abandoned when I went on vacation. I've put them on my to-do list to get to eventually, but it doesn't seem high enough priority to address immediately. Let me know if you'd find this useful somewhere else – we can change the priority.

PR #126, opened 2025-03-28
pathogen-repo-build: Fix extraction of AWS Batch job ids

tsibley:
Nextstrain CLI 9.0.0 added an improvement that automatically disables colorized and bolded output when stdout is not a terminal (or when NO_COLOR is set).  This broke our sed pattern which too-stringently expected ANSI escapes (matched by ".+") to surround the "AWS Batch Job ID:" label.

Bump our CLI requirement to 9.0.0 and remove expectation of ANSI escapes from the sed pattern.  We could handle both presence and absence of ANSI escapes, but its simpler to bump to 9.0.0 and expect absence.

In the future, we should stop parsing job logs and add a feature to Nextstrain CLI to directly emit this job id in a more structured way.

Related-to: <https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1743179523144269>

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/pathogen-repo-build.yaml
+++ b/.github/workflows/pathogen-repo-build.yaml
@@ -319,7 +319,7 @@ jobs:
       - name: Setup runtime ${{ inputs.runtime }}
         uses: ./.git/nextstrain/.github/actions/setup-nextstrain-cli
         with:
-          cli-version: ">=8.3.0"
+          cli-version: ">=9.0.0"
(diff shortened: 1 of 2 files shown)


victorlin (approved):

tsibley:
https://github.com/nextstrain/.github/pull/51#discussion_r1322114064 orz

